PR #15, opened 2019-09-02
Episode context

fshenton:
Episode component created and using Episodes context defined in the Episodes component folder. The activeIndex is not currently being updated but is accessible.

### DIFF
--- a/src/components/Episode/Episode.jsx
+++ b/src/components/Episode/Episode.jsx
@@ -1,15 +1,15 @@
 import React, { useContext } from "react";
-import { 
-	data, 
-	EpisodesContext as Episodes 
-} from "COMPONENTS/Episodes/";
+import { EpisodesContext as Episodes } from "COMPONENTS/Episodes/";
 import { s } from "./";
 
 export default function Episode(props){
 
 	//CONTEXT
 	//---------------------------
-	const { activeIndex } = useContext(Episodes);
+	const { 
+		activeIndex,
+		episodeCount: episodes
+	} = useContext(Episodes);
 
 
 	//RENDER
@@ -38,7 +38,7 @@ export default function Episode(props){
 			className={ s.wrapper }
 			role="group"
 			aria-roledescription="slide"
-			aria-label={ `Episode ${number} of ${data.length}.` }
+			aria-label={ `Episode ${number} of ${episodes}.` }
 			aria-hidden={ hidden.toString() }
 		>
 			<article>
--- a/src/components/Episode/Episode.jsx
+++ b/src/components/Episode/Episode.jsx
@@ -30,8 +30,7 @@ export default function Episode(props){
 	} = image;
 
 	const { videoSrc } = video;
-
-	const isActive = index === activeIndex;
+	const hidden       = index !== activeIndex;
 
 	return (
 		<li
@@ -40,7 +39,7 @@ export default function Episode(props){
 			role="group"
 			aria-roledescription="slide"
 			aria-label={ `Episode ${number} of ${data.length}.` }
-			aria-hidden={ !isActive.toString() }
+			aria-hidden={ hidden.toString() }
 		>
 			<article>
 				<h2>
--- a/src/components/Episode/Episode.jsx
+++ b/src/components/Episode/Episode.jsx
@@ -1,10 +1,15 @@
-import React from "react";
-import {
-	data as episodes
-} from "COMPONENTS/Episodes/";
+import React, { useContext } from "react";
+import { Episodes } from "COMPONENTS/Episodes/";
 
 export default function Episode(props){
 
+	//CONTEXT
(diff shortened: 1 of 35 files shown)


src/components/Episodes/Episodes.jsx
```diff
@@ -12,16 +13,18 @@ export default function Episodes(){
...
-				<button aria-label="Next episode." />
+			<EpisodesProvider>
 				<ol>
-					{ pagination }
+					{ episodes }
 				</ol>
-			</nav>
+				<nav>
+					<button aria-label="Previous episode." />
+					<button aria-label="Next episode." />
```
pookage: might be worth putting `aria-controls` on here and pointing to the episodes' `<ol>`

src/components/Episodes/Episodes.jsx
```diff
@@ -12,16 +13,18 @@ export default function Episodes(){
...
 	return (
 		<article>
-			<ol>
-				{ episodes }
-			</ol>
-			<nav>
-				<button aria-label="Previous episode." />
-				<button aria-label="Next episode." />
+			<EpisodesProvider>
 				<ol>
```
pookage: let's whack similar `aria` accessibility stuff on this to what we've done on the `<Characters>` carousel

src/components/Episodes/index.js
```diff
@@ -1,5 +1,6 @@
 import data from "./data.js";
 import RENDER from "./renderUtils.jsx";
+import { Episodes, EpisodesProvider } from "./context.jsx";
 
-export { data, RENDER };
+export { data, RENDER, Episodes, EpisodesProvider };
```
pookage: Just want to take a moment to acknowledge that this does kinda break my A/B naming-convention; this would, I guess be `EpisodesProvider` and `EpisodesContext` if that were to be followed - the only reason it's not is because, when we use it later, we use it as `useContext(Episodes)`, which reads better than `useContext(EpisodesContext)`.

THAT SAID, what we could do is:

```
export { data, RENDER, EpisodesContext, EpisodesProvider };
```

and then later...

```
import { EpisodesContext as Episodes } from "COMPONENTS/Episodes/";
```

what do you reckon?

fshenton: yeah I definitely prefer that way of doing it. Importing Episodes felt unclear to me previously, but bringing in EpisodeContext as Episodes makes a lot more sense, so we can useContext(Episodes) still. 👍

pookage: Cool - wanna refractor everywhere we've done this into this format then? Can mark the conversation as resolved when you're done! 🙌

fshenton:
Yeah will do. Just reinstalling node as conemu was having issues with npm
commands, despite environment path looking correct

src/components/Episodes/index.js
fshenton: Yeah will do. Just had to reinstall node (on laptop) as i couldn't use npm commands in terminal for some reason. Seems to be working ok now

src/components/Episodes/Episodes.jsx
fshenton: OK i've added the same accessibility attributes as we have in Character, all working fine except aria-hidden and I am very confused as to why there is an issue.

Using this gives us the correct result (episode 0 is active and not hidden, episode 1 is hidden):
- const flippedActive = !isActive;
- aria-hidden={ flippedActive.toString() }

Using aria-hidden={ !isActive.toString() } has them both set to false.

I just inspected our character list items and they're having the same problem. 
Any idea why the difference? I imagine there's some timing thing with doing the calculation (!) in the render return....hmmm

src/components/Episodes/Episodes.jsx
fshenton: Done. Used "episodes__items" as list id (see changes in nav branch).

src/components/Episode/Episode.jsx
```diff
@@ -21,8 +31,17 @@ export default function Episode(props){
...
+	const isActive = index === activeIndex;
+
 	return (
-		<li>
+		<li
+			id={ id }
+			className={ s.wrapper }
+			role="group"
+			aria-roledescription="slide"
+			aria-label={ `Episode ${number} of ${data.length}.` }
```
pookage: in hindsight this feels kinda weird to access the data directly - maybe we could include the number of episodes in the provider?

fshenton: Changed so we have an episodeCount property in our context provider. 
Worth noting that we could do the same thing for Characters, i chose not to as this branch is episodes only. Let me know if it's fine practice to make that change and I can.

pookage: feel free to make the changes on `markup` and just use the same implementation - maybe branch off, but no need to do a PR if the implementation is the same 👌

PR #16, opened 2019-09-02
Episode nav

fshenton:
Unclear why this pull request contains your Link refactor and my merges to markup?

The episode-nav branch should contain only the EpisodesControls changes as that's the only thing I committed to it. :worried:


pookage:
> Unclear why this pull request contains your Link refactor and my merges to markup?
> 
> The episode-nav branch should contain only the EpisodesControls changes as that's the only thing I committed to it. 😟

So, this will be because `episode-context` will have been branched from `markup` before my `link-refactor` branch was merged in; therefore that commit is new to this branch. It has the same commit hash as the one on `markup`, though, so those two commits won't double-up when this is merged into `develop`, don't worry 👌

src/components/Episodes/Episodes.jsx
```diff
@@ -8,23 +9,20 @@ import {
...
+				<EpisodesProvider>
 					<ol>
-						{ pagination }
+						{ episodes }
 					</ol>
-				</nav>
+				</EpisodesProvider>
+			</article>
+			<EpisodesProvider>
+				<EpisodesControls />
```
pookage: SUPER minor nitpick - could we name this component `<EpisodesNav />` instead? 🙏

src/components/Episodes/Episodes.jsx
```diff
@@ -8,23 +9,20 @@ import {
...
-			<EpisodesProvider>
-				<ol>
-					{ episodes }
-				</ol>
-				<nav>
-					<button aria-label="Previous episode." />
-					<button aria-label="Next episode." />
+		<>
+			<article>
+				<EpisodesProvider>
```
pookage: SO, we need `<EpisodesProvider>` to wrap _both_ the `episodes` _and_ the `<EpisodesControls>`, if we don't then we're creating two instances of the component, and thus there'll be two separate states with `activeIndex` being provided.

src/components/EpisodesControls/EpisodesControls.jsx
```diff
@@ -0,0 +1,25 @@
+import React, { useContext } from "react";
+import { data } from "COMPONENTS/Episodes/";
+import { RENDER } from "./";
+import { Episodes } from "COMPONENTS/Episodes/";
```
pookage: this should be imported from line 2 as you're already importing something from that component above...

src/components/EpisodesControls/EpisodesControls.jsx
```diff
@@ -0,0 +1,25 @@
...
+	const { activeIndex } = useContext(Episodes);
+
+	//RENDER
+	//---------------------------
+	const pagination = data.map(RENDER.pagination);
+
+	return (
+		<nav>
+			<button aria-label="Previous episode." />
+			<button aria-label="Next episode." />
```
pookage: in addition to my comment from the other branch (these buttons should have `aria-controls` for the `<ol>` of `episodes`, it occurrs to me that they also control the pagination below as well (ie. which one has `aria-current`)

src/components/EpisodesControls/renderUtils.jsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { s } from "./";
+import Link from "COMPONENTS/Link/";
+
+function renderPaginationLink(data){
+
+	const { number } = data;
+
+	return (
+		<li key={`episode__pagination__${number}`}>
```
pookage: I've just found out that [aria-current](https://www.w3.org/TR/wai-aria-1.1/#aria-current) exists! can we use that here? If we grab the `activeIndex` from the provider and compare it against the `index` here, then we should be able to apply that.

src/components/Episodes/Episodes.jsx
fshenton: Yarp. done

src/components/Episodes/Episodes.jsx
fshenton: Cool, had a feeling that would come back to me!
Wasn't sure about the nav being in the article, but it definitely seemed weird as sibling when it meant adding provider again.

src/components/EpisodesControls/EpisodesControls.jsx
fshenton: Done, however Episodes has been renamed in the context branch, so i'll want to update that. Sent you a PM about this as unsure the best way to sync that rename from context.

fshenton: Sorted.

src/components/EpisodesControls/EpisodesControls.jsx
fshenton: Added controls for episodes list.
TODO: using context to set aria-current and then looking at controls with that also.

src/components/EpisodesControls/renderUtils.jsx
fshenton: Now importing context and setting aria-current based on activeindex.

src/components/EpisodesControls/EpisodesControls.jsx
fshenton: Controls now point to the episode list and the pagination list.
Aria-current determined by activeindex.

src/components/Episodes/Episodes.jsx
```diff
@@ -8,32 +9,21 @@ import {
 
 export default function Episodes(){
 
-	const pagination = data.map(RENDER.pagination);
 	const episodes = data.map(RENDER.episode);
 
 	return (
+		<>
```
pookage: don't need a fragment if there's already a wrapper 👌

fshenton: Removed fragment!

PR #17, opened 2019-09-02
Episode preview

fshenton:


src/components/App/App.jsx
```diff
@@ -28,6 +29,7 @@ export default function App(){
 				<About />
 				<Characters />
 				<Episodes />
+				<EpisodePreview />
```
pookage: rather than having the `<EpisodePreview>` as another 'page', I think we discussed it as being a child of the `<Episode>` component, with the thumbnail triggering the full-screen mode. Something like:

```
<Episode>
    <h1>
        Episode 1: the whatever part 1
    </h1>
    <p>
        Synopsis
    </p>
    <EpisodePreview /> // the thumbnail
</Episode>
```

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,21 @@
...
+				/>
+				<video controls src="path/to/video.mp4" />
+			</output>
+			<h2>Episode X</h2>
+			<h1>Name of Episode X</h1>
+			<button>Previous</button>
+			<button>Next</button>
+		</form>
+	);
+}
```
pookage: A couple comments on the implementation with a `<form>` :

1. Try to group elements in the html where it makes logical sense; remember, at this stage we're almost not thinking about the styling at all - so for the 'close' button to appear by itself at the top is a bit weird; it should be at the bottom with the next / previous
2. If the next / previous control which is the current episode, then the headings should probably be in the `<output>` as well.

That said, as this would be a child of the `<Episode />` component, I think this works better if we come at it less as a form and more of an 'additional content' kinda thing. I'm thinking either:

- an `<aside>` with a kinda `<figure>`-like vibe, or
- a `<details>` - as we're using the thumbnail to hide / reveal the `<video>` and its associated content

if we go for the former, it'd be something like:

```
<aside>
	<a 
		href="path/to/episode/video.mp4"
		aria-controls="episode__1__preview"
	>
		<img src="path/to/episode/thumb.jpg" alt="thumb contents" />
	</a>
	<div 
		id="episode__1__preview"
		aria-hidden="true"
	>
		<header>
			<h1>
				Episode 1: The Whatever Part 1
			</h1>
			<nav>
				<a href="#episode-0" />
				<a href="#episode-2" />
			</nav>
		</header>
		<video>
			// only render the video sources when the figure is activated
		</video>
	</div>
</aside>
```
(only with a bunch more `aria` stuff)

and for the latter:

```
<details>
	<summary>
		<a href="path/to/episode/video.mp4">
			<img src="path/to/episode/thumb.jpg" alt="thumb contents" />
		</a>
	</summary>
	<div id="episode__1__preview">
		<div>
			<h1 id="episode__1__title">
				Episode 1: The Whatever Part 1
			</h1>
			<a href="#episode-0" />
			<a href="#episode-2" />
		</div>
		<video aria-labelledby="episode__1__title">
			// only render the video sources when the figure is activated
		</video>
	</div>
</details>
```

I think either would work, and not sure if that HTML is right, so worth having a ponder on it either way. Your call, cap'n!

fshenton: ok so i set it up to use the aside approach you outlined.

I have the thumbnail link toggling the hidden state for the preview overlay.
What's missing is the ability to turn on/off hidden for prev/next preview overlays, and updating the episode breakdown (that is obscured by the active overlay).
Thoughts on the best way to implement that?

Any other thoughts on it much appreciated. Bit confused 😄

src/components/App/App.jsx
fshenton: Done

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+					<nav>
+						{ prevLink }
+						{ nextLink }
+						<a href={ `#episode__${number}` }> 
+							Close
+						</a>
+					</nav>
+				</header>
+				<video>
+					{/*only render the video when it's activated*/}
```
pookage: We should add a `<source>` in here with `videoSrc` and only render it once activated (using the same method I demonstrated up there with the next / prev buttons ^^^)

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+		prevLink = 
+			<a href={ `#episode__${number-1}__preview` }>
+				Prev
+			</a>
+		;
+	}
+
+	if(nextIndex < episodeCount){
+		nextLink =
+			<a href={ `#episode__${number+1}__preview` }>
```
pookage: just spotted this - if `number` is an index then we should keep it called `index` to indicate that it's zero-indexed. If I saw `episode number 7` as a concept then I would assume it's the 7th, whereas `episode index 7` would be the 6th without needing to look anything up. If you wanna use `number` then the whole `+1` stuff should happen before that and have `const episodeNo = index + 1` or something.

desktop.ini
```diff
@@ -0,0 +1,4 @@
+[ViewState]
+Mode=
+Vid=
+FolderType=Generic
```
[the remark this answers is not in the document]
fshenton: No idea, likely just me dragging something in by accident. Still getting used to the trackpad haha. 
I've deleted it, seems fine.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+	//-----------------------------
+	const {
+		activeIndex,
+		episodeCount
+	} = useContext(Episodes);
+
+
+	// STATE
+	// ----------------------------
+	const [hidden, setHidden] = useState(true);
```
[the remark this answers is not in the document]
fshenton: OK. I ended up using the useState instead of the initialState object to pass the values and setter callbacks in with the context.provider. All working as expected:
- episode 0 is not hidden by default
- clicking thumbnail link sets the episode 0 preview hidden to be false
- clicking close or next sets preview hidden back to true
- prev/next changes the activeIndex and sets hidden to false for both the episode and the preview of that episode

The main thing remaining is having the prev/next anchors using video sources instead of anchor IDs. Struggling to find the best way to include index-1 source and index+1 source as part of the data mapping function in Episodes. Any thoughts on best approach? Will have another look tomorrow with fresh eyes

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+	// RENDER
+	// ----------------------------
+	const { 
+		index,
+		number,
+		title,
+		thumbSrc,
+		thumbDesc,
+		videoSrc
```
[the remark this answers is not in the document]
fshenton: Done 👍

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+		;
+	}
+
+
+	return (
+		<aside>
+			<a
+				href={ videoSrc }
+				aria-controls={ previewId }
+				onClick={ () => { setHidden(!hidden) } }
```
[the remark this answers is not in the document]
fshenton: Done.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+				Prev
+			</a>
+		;
+	}
+
+	if(nextIndex < episodeCount){
+		nextLink =
+			<a href={ `#episode__${number+1}__preview` }>
+				Next
+			</a>
```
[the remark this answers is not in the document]
fshenton: I knew you'd come up with something better, it did feel bad having the jsx declared above the render, but it also seemed overkill to do it in a renderUtils.

LOVE the isLast, isFirst booleans for selective rendering, AND glad I got to use the && syntax for it. Woop!

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+	const prevIndex = index - 1;
+	const nextIndex = index + 1;
+
+	let prevLink, nextLink;
+
+	if(prevIndex > -1){
+		prevLink = 
+			<a href={ `#episode__${number-1}__preview` }>
+				Prev
```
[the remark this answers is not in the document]
fshenton: Used aria-label as I like Aria <3

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: To clarify: number is not zero-indexed. 
Good feedback though. I've renamed as episodeNo in the destructure, set up 2 new vars called prevEpisodeNo and nextEpisodeNo and used them instead to make it cleaner.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+			</a>
+		;
+	}
+
+
+	return (
+		<aside>
+			<a
+				href={ videoSrc }
+				aria-controls={ previewId }
```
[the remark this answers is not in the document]
fshenton: Cheers, hadn't considered that but knew i was missing some accessibility stuff!
Added "Watch episode preview." as aria-label.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,93 @@
...
+	const { 
+		index,
+		number,
+		title,
+		thumbSrc,
+		thumbDesc,
+		videoSrc
+	} = props;
+
+	const previewId = `episode__${number}__preview`;
```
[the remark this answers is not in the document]
fshenton: Thanks for the reminder!

fshenton:
Got through a lot of the tougher stuff (at least I hope so!). Will look at video source, getting next/prev buttons in episodenav working (realised they weren't set up!) and updating hrefs of next/prev overlay buttons so they point at video sources.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+		setOverlayActive(show);
+	}// showOverlay
+
+	return (
+		<aside>
+			<a
+				href={ videoSrc }
+				aria-controls={ previewId }
+				aria-label="Watch episode preview."
```
pookage: I think we can improve this a liiitle more by doing something like `Watch preview for episode 1: The Awakening.`

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+				onClick={ (e) => { showOverlay(e, true) } }
+			>
+				<img
+					src={ thumbSrc }
+					alt={ thumbDesc }
+				/>
+			</a>
+			<div
+				id={ previewId }
+				aria-hidden={ hidden }
```
pookage: needs to be `hidden.toString()`

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+								onClick={ (e) => { 
+									handleNavPressed(e, index+1) 
+								}}
+							>
+								Next
+							</a>
+						)}
+						<a 
+							href={ `#episode__details__${episodeNo}` }
+							onClick={ e => showOverlay(e, false) }
```
pookage: So, more naming stuff - anything with a/b shouldn't take an argument - ie `showOverlay` implies the existence of `hideOverlay` and so you'd expect those functions to be called like `showOverlay()` and `hideOverlay()` - if it's one function that changes between two states then that'd be `toggleOverlay()`, but if it's based on a given argument then that's gotta be `setOverlay(true||false)`, right?

In this case, what I think we want is `setOverlay` and some binding to give us the `showOverlay` and `hideOverlay`.  BUT, we've already been given the terminlogy with the word `close` - therefore instead of `show`/`hide` we should be using `open`/`close` - ie:

```
function setOverlay(active, event){
    if(event) event.preventDefault();
    setOverlayActive(active);
}//setOverlay

const openOverlay = setOverlay.bind(this, true);
const closeOverlay = setOverlay.bind(this, false);

...

<a onClick={openOverlay}>
    Open
</a>

<a onClick={closeOverlay}>
   Close
</a>
```

The reason for `if(event) event.preventDefault()` is that it will use the `preventDefault()` when we use it for onClick events, but we are still able to use them as util-style functions with `setOverlay(false)` or even just `openOverlay()`.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+		setActiveIndex(newIndex);
+	}// handleNavPressed
+
+
+	function showOverlay(e, show){
+		e.preventDefault();
+
+		setOverlayActive(show);
+	}// showOverlay
```
pookage: notes on this further down in another comment.

src/components/Episodes/context.jsx
```diff
@@ -1,23 +1,29 @@
...
-	activeIndex: 0,
-	episodeCount: episodes
-};
 
 function EpisodesProvider(props){
 	const { children } = props;
 
+	const episodeCount = data.length;
+	const [activeIndex, setActiveIndex] = useState(0);
+	const [overlayActive, setOverlayActive] = useState(false);
```
pookage: Let's restore the `initialState` object and use the values from that as the default values for these two `useState` calls

src/components/Episodes/context.jsx
```diff
@@ -1,23 +1,29 @@
...
+
+	const state = {
+		episodeCount,
+		activeIndex,
+		setActiveIndex,
+		overlayActive,
+		setOverlayActive
```
pookage: as mentioned earlier - let's move these into two child `get` and `set` objects for values / functions

src/components/Episode/Episode.jsx
```diff
@@ -51,16 +46,13 @@ export default function Episode(props){
...
-						src={ thumbSrc } 
-						alt={ thumbDesc } 
-					/>
-				</a>
+				<EpisodePreview 
+					index={ index }
+					number={ number }
+					title={ title }
+					image={ image }
+					video={ video }
```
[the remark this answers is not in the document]
pookage: Also, want to come up with some kinda convention for prop-order? even if it's just alphabetical?

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Shweet, done.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+export default function EpisodePreview(props){
+
+	// CONTEXT
+	//-----------------------------
+	const {
+		episodeCount, 
+		activeIndex,
+		setActiveIndex,
+		overlayActive,
+		setOverlayActive
```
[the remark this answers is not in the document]
fshenton: Decided against destruring get and set from context, so we could do set.activeIndex etc., as I find the longer form (episodes.set.activeIndex) to be more descriptive. Makes the code lines fairly long though, so happy to change that.

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Good shout, done.

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Yaaas.

src/components/Episode/Episode.jsx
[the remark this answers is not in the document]
fshenton: Updated to use thumbnail and trailer throughout.

Current convention (id's/indexs/nums -> text content -> media):
id={ id }
key={ id }
index={ index }
episodeNo={ episodeNo }
title={ title }
synopsis={ synopsis }
thumbnail={ thumbnail }
trailer={ trailer }

Let me know if you've got a better suggestion <3

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+	// RENDER
+	// ----------------------------
+	const { 
+		index,
+		number: episodeNo,
+		title,
+		image: {
+			src: thumbSrc,
+			description: thumbDesc
```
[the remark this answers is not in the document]
fshenton: Done

src/components/Episode/Episode.jsx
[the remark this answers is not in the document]
fshenton: (happy to do alphabetical if that's a better standard to follow!)

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Cool yeah, makes sense.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,120 @@
...
+		},
+		video: {
+			src: videoSrc
+		}
+	} = props;
+
+	const previewId = `episode__${episodeNo}__preview`;
```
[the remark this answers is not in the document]
fshenton: Updated id so it's more readable in URL.

Not sure the best way to approach the prev/next episode IDs though! 🤔

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+	const previewId = title
+		.replace(/\W+/g, '-') // remove non-alphanumeric & replace with -s
+		.toLowerCase()
+		+"-preview"
+		;
+
+	const prevPreviewId = `episode__${prevEpisodeNo}__preview`;
+	const nextPreviewId = `episode__${nextEpisodeNo}__preview`;
+
+	function handleNavPressed(e, newIndex){
```
pookage: This should be `setActiveIndex` or something - `handleSomething` is used when the function is doing so many things that you can't really give it a useful name and so you're just like IT'S THE THING THAT HAPPENS WHEN YOU CLICK IT.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+								}}
+							>
+								Prev
+							</a>
+						)}
+						{ !isLast && (
+							<a 
+								href={ nextPreviewId } 
+								onClick={ (e) => { 
+									handleNavPressed(e, index+1)
```
pookage: This should be `onClick={next}` done by going `const next = handleNavPressed.bind(this, index+1);`

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+								onClick={ (e) => { 
+									handleNavPressed(e, index+1) 
+								}}
+							>
+								Next
+							</a>
+						)}
+						<a 
+							href={ episodeId }
+							onClick={ e => closeOverlay(e) }
```
pookage: as above - this doesn't need an arrow function - calling `onClick={closeOverlay}` will do exactly the same thing as this.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+			description: thumbDesc
+		},
+		trailer: {
+			src: trailerSrc
+		}
+	} = props;
+
+	// handle visibility of active episode and episode preview overlay
+	const isActiveEpisode 	= episodes.get.activeIndex === index;
+	const thisOverlayActive = isActiveEpisode && episodes.get.overlayActive;
```
pookage: `this` is a very loaded word, so we shouldn't use it in variable names otherwise it just looks like we're missing a `.` (ie. `this.overlayActive` or something) - also, you've set a precedent for naming in the line above with `isActiveEpisode` and so we might as well continue it with `isOverlayActive`.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+	const isLast 	= index === episodes.get.episodeCount - 1;
+
+	const prevEpisodeNo = episodeNo - 1;
+	const nextEpisodeNo = episodeNo + 1;
+
+	const previewId = title
+		.replace(/\W+/g, '-') // remove non-alphanumeric & replace with -s
+		.toLowerCase()
+		+"-preview"
+		;
```
pookage: You're doing two things here:
1. making it url-safe
2. appending 'preview'. 

I would say that a better way to do this would be to change the `generateSafeString` in the utils so that you can optionally pass in a second argument to determine whether you use a `_` or a `-` to replace words, and then do:

```
const previewId = generateSafeString(`${title} preview`, "-");
```
or
```
const previewId = `${generateSafeString(title, "-")}-preview`;
```

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+							Close
+						</a>
+					</nav>
+				</header>
+				{ thisOverlayActive && 
+					<video
+						src={ trailerSrc }
+					>
+					</video>
+				}
```
pookage: Instead of preventing the render of the whole video element, can we just do it for its sources? ie.

```
<video 
    aria-hidden={isOverlayActive}
    aria-label={`Episode ${episodeNo} preview.`}
>
    {isOverlayActive && (
        <source src={trailerSrc} type="video/mp4" />
    )
</video>
```

That way the video element is THERE and we can add whatever controls we want to it etc etc later, but we just don't cause the heavy loading of video files until we need to.  It would also be a bit odd not have something here, otherwise we would have a `<header>` (introductory content) but no actual content for it to be introducing

src/components/Episodes/data.js
```diff
@@ -1,22 +1,22 @@
 export default [
 	{
-		number: 1,
-		image: {
+		episodeNo: 1,
```
pookage: the values here should still be `number` not `episodeNo` - it's in a data object that represents an array of episodes, so it's clear that the number would be the episode number; `episodeNo` is just what we can name them during the extraction process or whatever.

src/components/Episodes/renderUtils.jsx
```diff
@@ -4,25 +4,28 @@ import Episode from "COMPONENTS/Episode/";
 function renderEpisode(data, index){
 
 	const { 
-		number,
+		episodeNo,
```
pookage: `number: episodeNo` as per the above.

src/components/Episodes/renderUtils.jsx
```diff
@@ -4,25 +4,28 @@ import Episode from "COMPONENTS/Episode/";
...
-		video,
+		thumbnail,
+		trailer,
 	} = data;
 
-	const id = `episode__${number}`;
+	const id = title
+			.replace(/\W+/g, '-') // remove non-alphanumeric & replace with -s
+			.toLowerCase()
+			;
```
pookage: Use the same util we mentioned in the other component

src/components/Episodes/renderUtils.jsx
```diff
@@ -4,25 +4,28 @@ import Episode from "COMPONENTS/Episode/";
...
 
 	return (
 		<Episode
 			id={ id }
 			key={ id }
 			index={ index }
-			number={ number }
+			episodeNo={ episodeNo }
```
pookage: ah, we might not even need to alias out of the data-object, as we're giving them straight as props, which will alias them from here on out - we can just do `episodeNo={number}` here if we wanted to.

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -11,20 +11,20 @@ function renderPaginationLink(data, index){
 
 	// RENDER
 	// --------------------------
-	const { number } = data;
+	const { episodeNo } = data;
```
pookage: don't forget to update this once the data-object has been udpated!

src/components/Episodes/renderUtils.jsx
fshenton: Created convertToHyphenatedString in utils.js, now being imported and used. 
Name is lengthy but descriptive :)

src/components/Episodes/renderUtils.jsx
fshenton: Changed back to number={number}. Then destructuring out as episodeNo in EpisodePreview.  Makes more sense in there as there will be a reference to prev and next episodes. Depending on implementation of prev/next episode ids (for next and previous with overlay open), might need to revisit this.

src/components/EpisodesNav/renderUtils.jsx
fshenton: Created episode-nav-state branch off episode-preview to do this. Will merge down once done.

src/components/Episode/Episode.jsx
```diff
@@ -1,14 +1,17 @@
...
 	//---------------------------
-	const { 
-		activeIndex,
-		episodeCount: episodes
+	const {
+		get: {
+			episodeCount,
+			activeIndex
+		}
 	} = useContext(Episodes);
```
pookage: Might as well do:
```
const { 
    episodeCount,
    activeIndex
} = useContext(Episodes).get;
```
here, as we're not pulling-out anything other than `get` from the context.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,134 @@
...
+			description: thumbDesc
+		},
+		trailer: {
+			src: trailerSrc
+		}
+	} = props;
+
+	// handle visibility of active episode and episode preview overlay
+	const isActiveEpisode 	= activeIndex === index;
+	const thisOverlayActive = isActiveEpisode && overlayActive;
```
pookage: oop - think I mentioend this before; this should be `const isOverlayActive`

src/shared/utils.js
```diff
@@ -4,6 +4,12 @@ function convertToSafeString(string){
 			.toLowerCase();
 }//convertToSafeString
 
+function convertToHyphenatedString(string){
+	return string.replace(/\W+/g, '-') // remove non-alphanumeric & replace with -s
+			.toLowerCase()
+}//convertToHyphenatedString
```
pookage: I don't think we need a separate function for this. If we just change `convertToSafeString` to be:
```
function convertToSafeString(string, separator="_"){
    return string.replace(/\W/g, "").replace(/ /g, separator).toLowerCase();
}
```
then we can just do convertToSafeString("whatever episode #2", "-"); if we want to separate with hyphens instead of underscores.

src/components/Episodes/renderUtils.jsx
```diff
@@ -1,17 +1,18 @@
...
 		title,
 		synopsis,
-		image,
-		video,
+		thumbnail,
+		trailer,
 	} = data;
 
-	const id = `episode__${number}`;
+	const id = UTILS.convertToHyphenatedString(title);
```
pookage: see my note in `utils` about why we should use `convertToSafeString` here instead.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,134 @@
...
+	const hidden 			= !thisOverlayActive;
+
+	const isFirst 	= index === 0;
+	const isLast 	= index === episodeCount - 1;
+
+	const prevEpisodeNo = episodeNo - 1;
+	const nextEpisodeNo = episodeNo + 1;
+
+	const previewId = UTILS.convertToHyphenatedString(title) + "-preview";
```
pookage: See my note in `utils.js`

src/components/EpisodePreview/EpisodePreview.jsx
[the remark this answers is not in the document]
fshenton: Sorted prev/next links. Now they point at prev/next video.

src/components/EpisodesNav/renderUtils.jsx
fshenton: All working and merged back to episode-preview.

src/components/Episode/Episode.jsx
fshenton: Done.

src/shared/utils.js
fshenton: Done

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: good spot, cheers

fshenton:
OK, fingers crossed that's everything :)

src/components/Episode/Episode.jsx
```diff
@@ -51,16 +48,16 @@ export default function Episode(props){
...
-				</a>
+				<EpisodePreview 
+					episodeId={ id }
+					index={ index }
+					number={ number }
+					title={ title }
+					thumbnail={ thumbnail }
+					trailer={ trailer }
+					prevTrailer={ prevTrailer }
+					nextTrailer={ nextTrailer }
```
pookage: mmm, not sure if I like this component receiving props JUST to pass them down to a child component - seems like the sort of thing that `context` would be used for? Let's do that - in the `renderEpisode` util, let's do an `<Episode.Provider value={{ prevTrailer, nextTrailer }}>` so that `<EpisodePreview>` can just do `const { prevTrailer, nextTrailer } = useContext(Episode);` - no need for a big old reducer or anything as it doesn't need any setters - it's just providing context without the need to pass props down multiple levels...

pookage: might need to create a `context.jsx` and `<EpisodeProvider>` for this? not sure.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,140 @@
...
+	const isOverlayActive   = isActiveEpisode && overlayActive;
+	const hidden 			= !isOverlayActive;
+
+	const isFirst = index === 0;
+	const isLast  = index === episodeCount - 1;
+
+	const prevEpisodeNo = episodeNo - 1;
+	const nextEpisodeNo = episodeNo + 1;
+
+	const previewId = UTILS.convertToSafeString(title, "-") + "-preview";
```
pookage: can we use a template literal for this? string concatenation just looks hella gross.

src/components/EpisodePreview/EpisodePreview.jsx
pookage: don't forget about this!

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,128 @@
...
+					<h1>
+						{ title }
+					</h1>
+					<nav>
+						{ !isFirst && (
+							<a 
+								href={ prevPreviewId }
+								aria-label="Previous."
+								onClick={ (e) => { 
+									handleNavPressed(e, index-1)
```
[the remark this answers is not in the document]
pookage: don't forget about this!

src/components/EpisodePreview/EpisodePreview.jsx
pookage: ...or this!

src/components/Episodes/Episodes.jsx
```diff
@@ -9,7 +9,18 @@ import {
...
+		//passing prev and next video src for video nav buttons
+		const prev = index !== 0 
+			? data[index-1] 
+			: {};
+		
+		const next = index !== data.length-1 
+			? data[index+1] 
+			: {};
+
+		return RENDER.episode(obj, index, prev, next);
```
pookage: I think we should just give `RENDER.episode` `data[index-1]` and `data[index+1]`  and then the logic relating to handling whether or not next/prev is `undefined` is just handled in the `<Episode>` component, or whatever.

Aside from that, this sorta logic should be part of the `RENDER.episode` util, as it's part of the episode stuff, not the episodes render itself, if that makes sense?

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -3,21 +3,58 @@ import {
...
+	// -------------------------------
+	const isFirstEpisodeActive = activeIndex === 0;
+	const isLastEpisodeActive  = activeIndex === episodeCount - 1;
+
 	const pagination = data.map(RENDER.pagination);
+	
+
+	// CLICK HANDLER
+	// -------------------------------
+	function handleNavPressed(e, newIndex){
```
pookage: see my earlier comments re: `handleWhatever` function names - this should be `setIndex` or something

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -3,21 +3,58 @@ import {
...
 	return (
 		<nav>
 			<button 
+				className={ s.navButton }
 				aria-label="Previous episode."
 				aria-controls="episodes__items episodes__pagination"
+				aria-hidden={ isFirstEpisodeActive }
+				onClick={ (e) => { 
+					handleNavPressed(e, activeIndex - 1) 
+				}}
```
pookage: as before - this needs to to be `const previous = setIndex.bind(this, activeIndex-1);` and `onClick={previous}` - hang in there, man. Almost there!

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -3,21 +3,58 @@ import {
...
+		setActiveIndex(newIndex);
+	}// handleNavPressed
 
 	return (
 		<nav>
 			<button 
+				className={ s.navButton }
 				aria-label="Previous episode."
 				aria-controls="episodes__items episodes__pagination"
+				aria-hidden={ isFirstEpisodeActive }
```
pookage: remember to `.toString()` this or it'll be weird and won't work.

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -3,21 +3,58 @@ import {
...
+				}}
 			/>
 			<button 
+				className={ s.navButton }
 				aria-label="Next episode." 
 				aria-controls="episodes__items episodes__pagination"
+				aria-hidden={ isLastEpisodeActive }
+				onClick={ (e) => { 
+					handleNavPressed(e, activeIndex + 1) 
+				}}
```
pookage: `const next = setIndex.bind(this, activeIndex+1);` `onClick={next}`

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -1,30 +1,52 @@
...
+		title
+	} = data;
+
+	const episodeId = UTILS.convertToSafeString(title, "-");
 
 	const isActive = index === activeIndex;
 
+	// CLICK HANDLER
+	// --------------------------
+	function handlePageClick(){
```
pookage: change this to a better function name - `handlePageClick` isn't much better than `doTheThing()` ya know?

src/components/EpisodesNav/styles.scss
```diff
@@ -1,3 +1,14 @@
 .paginationLink {
 	font-size: 1em; //placeholder to have the classname showing in inspector
+}
+
+.navButton {
+	width: 50px;	//placeholder so they are visible
+	height: 50px;
+	margin: 5px;
+
+	&[aria-hidden=true] {
```
pookage: does this work? I think you need `&[aria-hidden="true"]`. I might be wrong, though.

src/shared/utils.js
```diff
@@ -1,6 +1,6 @@
-function convertToSafeString(string){
-	return string.replace(/\W/g, '') // remove non-alphanumeric
-			.replace(/ /g, "_")      // replace spaces with underscores
+function convertToSafeString(string, seperator="_"){
```
pookage: oh, btw, if you haven't seen this yet - this is a way to specify a default value for function arguments - in the same way as if we were doing object destructuring.

src/components/Episode/Episode.jsx
fshenton: Created a TrailersProvider which is now passing prevTrailer and nextTrailer objects, which are in turn destructured inside EpisodePreview for their sources.

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Done.

src/components/EpisodesNav/EpisodesNav.jsx
fshenton: As discussed, we don't need to due to React magic!

src/components/EpisodesNav/styles.scss
fshenton: It works, but as discussed not best practice.

src/components/EpisodesNav/EpisodesNav.jsx
fshenton: done

src/components/EpisodesNav/renderUtils.jsx
fshenton: Done.

src/shared/utils.js
fshenton: old news mate 😏

fshenton:
Right, that should actually be everything 👀

src/components/Episodes/actions.js
```diff
@@ -0,0 +1,11 @@
...
+export const NEXT_EPISODE 			= "NEXT_EPISODE";
+export const GET_EPISODE			= "GET_EPISODE";
+export const SHOW_TRAILER			= "SHOW_TRAILER";
+
+export default {
+	PREV_EPISODE,
+	NEXT_EPISODE,
+	GET_EPISODE,
+	SHOW_TRAILER
+};
```
pookage: Wait, isn't `export default` going to cause a problem with the whole `import * as ACTIONS` thing in `index.js`? Or is that in the other branch? If it's in the other branch then you need to remember about this during the merge as some stuff might be overwritten - best to avoid refactoring the same file on multiple simultaneous branches or you're going to encounter issues like this one.  Unfortunately you can't go back and remove files from a commit, so it's just something you need to be mindful about.

src/components/Episodes/context.jsx
```diff
@@ -1,23 +1,58 @@
...
 			{ children }
 		</EpisodesContext.Provider>
 	);
-}
+}// EpisodesProvider
+
+
+// TRAILER CONTEXT
+// ------------------------------
+const TrailersContext = createContext();
```
pookage: I think this context should be in `Episode/context.jsx` as it's going to be part of the individual episode render, if ya get me?

src/components/Episodes/data.js
```diff
@@ -1,22 +1,22 @@
 export default [
 	{
 		number: 1,
-		image: {
+		thumbnail: {
```
pookage: didn't we already do this? is this a duplication of work from the other branch?

src/components/Episodes/index.js
```diff
@@ -1,6 +1,8 @@
 import data from "./data.js";
+import ACTIONS from "./actions.js";
+import reducer from "./reducer.jsx";
 import RENDER from "./renderUtils.jsx";
-import { EpisodesContext, EpisodesProvider } from "./context.jsx";
+import { EpisodesContext, EpisodesProvider, TrailersContext, TrailersProvider } from "./context.jsx";
```
pookage: remember to remove when you move it to `Episode/context.jsx`

src/components/Episodes/index.js
```diff
@@ -1,6 +1,8 @@
...
 
-export { data, RENDER, EpisodesContext, EpisodesProvider };
+export { data, RENDER, reducer, ACTIONS, EpisodesContext, EpisodesProvider, TrailersContext, TrailersProvider };
```
pookage: remember to remove when you move it to `Episode/context.jsx`

src/components/Episodes/context.jsx
```diff
@@ -1,23 +1,58 @@
...
+
+function TrailersProvider(props){
+	const {
+		children,
+		value: {
+			prevTrailer,
+			nextTrailer
```
pookage: could we have these as `next` and `previous` props instead of a prop called `value` with them inside?

src/components/Episodes/Episodes.jsx
```diff
@@ -9,7 +9,18 @@ import {
 
 export default function Episodes(){
 
-	const episodes = data.map(RENDER.episode);
+	const episodes = data.map( (obj, index) => {
+		//passing prev and next video src for video nav buttons
+		const prevTrailer = index !== 0 
+			? data[index-1].trailer
+			: {};
```
pookage: Did we fix this on the other branch? Deffo shouldn't be duplicating work if so but if not then this should done in the render util

src/components/Episodes/Episodes.jsx
```diff
@@ -9,7 +9,18 @@ import {
...
+		
+		const nextTrailer = index !== data.length-1 
+			? data[index+1].trailer
+			: {};
```
pookage: see above.

src/components/Episodes/Episodes.jsx
```diff
@@ -9,7 +9,18 @@ import {
 
 export default function Episodes(){
 
-	const episodes = data.map(RENDER.episode);
+	const episodes = data.map( (obj, index) => {
```
pookage: only just realised this was an arrow inside a `.map` - I swear we fixed this - was it another branch? Can we restore this to `data.map(RENDER.episode)` and just do all of this logic inside that util? It's what the fellah is there for 🙏

src/components/Episodes/renderUtils.jsx
```diff
@@ -1,29 +1,36 @@
 import React from "react";
 import Episode from "COMPONENTS/Episode/";
+import { TrailersProvider } from "./";
+import UTILS from "SHARED/utils.js";
 
-function renderEpisode(data, index){
+function renderEpisode(data, index, prevTrailer, nextTrailer){
```
pookage: needs to be changed back to `function renderEpisode(data, index){`

src/components/Episodes/renderUtils.jsx
```diff
@@ -1,29 +1,36 @@
...
-			id={ id }
-			key={ id }
-			index={ index }
-			number={ number }
-			title={ title }
-			synopsis={ synopsis }
-			image={ image }
-			video={ video }
-		/>
+		<TrailersProvider
```
pookage: Yeah, the context definintely needs to be `Trailer` then, otherwise it sounds like we're providing information for more than one trailer - so this needs to be `<TrailerProvider>`

src/components/Episodes/renderUtils.jsx
```diff
@@ -1,29 +1,36 @@
...
-			key={ id }
-			index={ index }
-			number={ number }
-			title={ title }
-			synopsis={ synopsis }
-			image={ image }
-			video={ video }
-		/>
+		<TrailersProvider 
+			value={ { prevTrailer, nextTrailer } }
```
pookage: could we have `next` and `previous` as the props instead of `value`?

pookage: and rather than them being passed down as arguments in this `renderEpisode` function, it should just be:

```

const { trailer: nextTrailer } = data[index+1] || {};
const { trailer: prevTrailer } = data[index-1] || {};

<TrailerProvider
    next={nextTrailer}
    previous={prevTrailer}
>
    <Episode />
</TrailerProvider>
```

src/components/Episodes/Episodes.jsx
```diff
@@ -9,7 +9,18 @@ import {
...
+		//passing prev and next video src for video nav buttons
+		const prevTrailer = index !== 0 
+			? data[index-1].trailer
+			: {};
+		
+		const nextTrailer = index !== data.length-1 
+			? data[index+1].trailer
+			: {};
+
+		return RENDER.episode(obj, index, prevTrailer, nextTrailer);
```
pookage: yerr, kill this whole arrow function - let's just use the render util - the point of it is preciseley to avoid doing all of this logic in the main component function.

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -1,23 +1,61 @@
...
+	// -------------------------------
+	const isFirstEpisodeActive = activeIndex === 0;
+	const isLastEpisodeActive  = activeIndex === episodeCount - 1;
+
 	const pagination = data.map(RENDER.pagination);
+	
+
+	// CLICK HANDLER
+	// -------------------------------
+	function changeEpisode(newIndex, e){
```
pookage: this can be `setActiveIndex` now 👍

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -1,30 +1,54 @@
...
+		title
+	} = data;
+
+	const episodeId = UTILS.convertToSafeString(title, "-");
 
 	const isActive = index === activeIndex;
 
+	// CLICK HANDLER
+	// --------------------------
+	function changeEpisode(){
```
pookage: this can be `setActiveIndex`

src/components/Episode/Episode.jsx
fshenton: Oops, totally forgot to remove those references 👀

src/components/Episode/Episode.jsx
```diff
@@ -20,26 +21,22 @@ export default function Episode(props){
 		number,
 		title,
 		synopsis,
-		image = {},
-		video = {},
+		thumbnail = {},
+		trailer = {},
+		prevTrailer = {},
+		nextTrailer = {}
```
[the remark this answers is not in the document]
fshenton: yarp, sorry, somehow missed this.

src/components/Episodes/Episodes.jsx
fshenton: Nah, it's been like that for a bit.
Some of the changes requested are hidden for some reason, so might be that I missed it in previous pr.

src/components/Episodes/data.js
fshenton: err we did this ages days ago 🤔

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,147 @@
...
+	// handle visibility of active episode and episode preview overlay
+	const isActiveEpisode 	= activeIndex === index;
+	const isOverlayActive   = isActiveEpisode && overlayActive;
+	const hidden 			= !isOverlayActive;
+
+	const isFirst 			= index === 0;
+	const isLast  			= index === episodeCount - 1;
+
+	const formattedTitle 	= UTILS.convertToSafeString(title, "-");
+	const previewId 	 	= `${formattedTitle}-preview`;
```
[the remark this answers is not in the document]
fshenton: Looks good on my screen 😕 
image

src/components/Episodes/actions.js
fshenton: It's working fine like this. 
The ACTIONS import = an object containing the different actions.
Could still import them seperately if needed, and then use without the ACTIONS namespace (when importing into a component).

src/components/Episode/index.js
```diff
@@ -1,4 +1,6 @@
 import s from "./styles.scss";
+import { TrailerContext, TrailerProvider } from "./context.jsx";
 
 export { s };
+export { TrailerContext, TrailerProvider };
```
pookage:
```suggestion
export { s, TrailerContext, TrailerProvider };
```

src/components/Episode/index.js
```diff
@@ -1,4 +1,6 @@
 import s from "./styles.scss";
+import { TrailerContext, TrailerProvider } from "./context.jsx";
 
 export { s };
```
pookage:
```suggestion
```

pookage: kill this line and merge with the contexts below.

src/components/EpisodePreview/EpisodePreview.jsx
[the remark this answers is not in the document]
pookage: ahhh, I just checked and it's because you've mixed tabs & spaces in the wrong way

- tabs for indentation
- spaces for formatting

image

By using tabs in formatting, the formatting goes wild whenever you change your tab-width. Your default is the same as mine- namely 4 spaces, but if I set my tab-width to 2 spaces...

image

So yeah - tabs to indent; spaces for formatting.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,143 @@
...
+
+
+	// CLICK HANDLERS
+	// -----------------------------
+	function setActiveIndex(newIndex, e){
+		e.preventDefault();
```
pookage: could we add if(e) here as well?

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,143 @@
...
+			episodeCount
+		},
+		dispatch
+	} = useContext(Episodes);
+
+
+	const trailers = useContext(Trailer);
+
+	const { src: prevTrailerSrc } = trailers.prev || {};
+	const { src: nextTrailerSrc } = trailers.next || {};
```
pookage: Let's do:

```
const { 
    prev: prevTrailer = {}, 
    next: nextTrailer = {}
} = useContext(Trailer);

const { src: prevTrailerSrc } = prevTrailer;
const { src: nextTrailerSrc } = nextTrailer;
```

That way we can use the whole 'default' value stuff that comes with destructuring instead of the `|| {}` fallback, which is less explicity 'default'.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,143 @@
...
+		e.preventDefault();
+
+		dispatch({
+			type: ACTIONS.GET_EPISODE,
+			value: newIndex
+		});
+	}// changeEpisode
+
+	const prevEpisode  = setActiveIndex.bind(this, index - 1);
+	const nextEpisode  = setActiveIndex.bind(this, index + 1);
```
pookage: Let's move all these click handlers & binding up above the render section, but also put the two functions together and all the binding together. I get the desire to group variables with their functions, but it's better to group all the functions together in one place than have to go hunt for them.

So the component structure would be:

```
1. HOOKS
2. FUNCTION DECLARATIONS
3. FUNCTION BINDINGS
4. RENDER LOGIC
5. RETURN
```

src/components/Episodes/index.js
```diff
@@ -1,6 +1,8 @@
 import data from "./data.js";
+import ACTIONS from "./actions.js";
```
pookage: ahh, did the `import * as ACTIONS from "./actions.js"` not work after all? </3

src/components/Episodes/index.js
```diff
@@ -1,6 +1,8 @@
...
+import reducer from "./reducer.jsx";
 import RENDER from "./renderUtils.jsx";
 import { EpisodesContext, EpisodesProvider } from "./context.jsx";
 
-export { data, RENDER, EpisodesContext, EpisodesProvider };
+export { data, RENDER, reducer, ACTIONS, EpisodesContext, EpisodesProvider };
```
pookage: not for now, but for another time, I'm thinking we should maybe just export `Context` and `Provider`, and then we can do:

```
import { Context as EpisodesContext } from "COMPONENTS/Episodes/";
```

Everything else is pretty neatly generic now, so it feels strange that this one's not! Super low priority, though, I might just do that on a branch myself at some point this weekend - I've already forced you to arbitrarily rename enough things 😅

src/components/Episodes/renderUtils.jsx
```diff
@@ -1,29 +1,40 @@
...
-			number={ number }
-			title={ title }
-			synopsis={ synopsis }
-			image={ image }
-			video={ video }
-		/>
+		<TrailerProvider 
+			prev={ prevTrailer }
+			next={ nextTrailer } 
+			key={ `${id}-trailer-provider` }
```
pookage: BEM! `provider__trailer__${id}`

pookage: well, most bem-iest bem would be `provider__trailer--${id}` but I tend to save that for CSS only - I keep to either `_` or `-` for ids/keys/classes, with my arbitrary rule being that if it's for internal logic, use `_` and if the user's going to see it then use `-`.

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -1,23 +1,61 @@
...
+		dispatch
+	} = useContext(Episodes);
+
+
+	// RENDER
+	// -------------------------------
+	const isFirstEpisodeActive = activeIndex === 0;
+	const isLastEpisodeActive  = activeIndex === episodeCount - 1;
+
 	const pagination = data.map(RENDER.pagination);
```
pookage: The render should be the absolute last thing before the `return` - everything else is basically getting variables and values ready for this section, so there shouldn't be anything under it.

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -1,23 +1,61 @@
...
+		e.preventDefault();
+
+		dispatch({
+			type: ACTIONS.GET_EPISODE,
+			value: newIndex
+		});
+	}// changeEpisode
+
+	const prevEpisode = setActiveIndex.bind(this, activeIndex - 1);
+	const nextEpisode = setActiveIndex.bind(this, activeIndex + 1);
```
pookage: Move this click-handler section up above the render section 👍

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -1,30 +1,54 @@
...
 	// --------------------------
-	const { number } = data;
+	const { 
+		number: episodeNo,
+		title
+	} = data;
+
+	const episodeId = UTILS.convertToSafeString(title, "-");
 
 	const isActive = index === activeIndex;
```
pookage: as mentioned above - make sure that 'render' is the last section before the `return` - so move the click-handlers up above it.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,143 @@
...
+			src: trailerSrc
+		}
+	} = props;
+
+	// handle visibility of active episode and episode preview overlay
+	const isActiveEpisode 	= activeIndex === index;
+	const isOverlayActive   = isActiveEpisode && overlayActive;
+	const hidden 			= !isOverlayActive;
+
+	const trailerSrcAttr    = isOverlayActive ? {source: trailerSrc} : {};
```
pookage: I get what you're trying to do here, but instead of having `<video src={src}>` we should have
```
<video>
    <source src={src} type="video/mp4" />
</video>
```

and because we're using a `<source>` element, we can just do the conditional rendering down in the return.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -0,0 +1,143 @@
...
+							onClick={ closeOverlay }
+						> 
+							Close
+						</a>
+					</nav>
+				</header>
+				<video
+					{ ...trailerSrcAttr } 
+				>
+				</video>
```
pookage: as mentioned above, can we instead make this:
```
<video>
    {isOverlayActive && (
        <source src={trailerSrc} type="video/mp4" />
    )}
</video>

src/components/Episode/index.js
pookage: just found out I could make suggestions like this! looks like I can only make single-line suggestions though...

PR #20, opened 2019-09-10
Episodes reducer

fshenton:
Ugh, this was a bit of a headache to get working. One of those slightly infuriating times where it resolved itself for no obvious reason. Might have been the way I was exporting and importing.

Works with importing the named actions, rather than importing actions and using actions.NAME.

Interested to see if you can spot where things might be going weird! I was getting absolutely bizarre console.logs while trying to debug...


src/components/Episodes/context.jsx
```diff
@@ -1,35 +1,25 @@
...
-		},
-		set: {
-			activeIndex: setActiveIndex,
-			overlayActive: setOverlayActive 
-		}
-	};
- 
 	return (
-		<EpisodesContext.Provider value={ state }>
+		<EpisodesContext.Provider value={ {state, dispatch} }>
```
pookage: super minor - for formatting I would do either `{{ state, dispatch }}` or `{ { state, dispatch } }` here.

src/components/Episodes/index.js
```diff
@@ -1,6 +1,8 @@
 import data from "./data.js";
+import { PREV_EPISODE, NEXT_EPISODE, SET_ACTIVE_INDEX, SET_OVERLAY_ACTIVE } from "./actions.js";
```
pookage: ahhh, is this what you were talking about? I _think_ you can do `import * as ACTIONS from "./actions.js";` here.

src/components/Episodes/index.js
```diff
@@ -1,6 +1,8 @@
...
+import reducer from "./reducer.jsx";
 import RENDER from "./renderUtils.jsx";
 import { EpisodesContext, EpisodesProvider } from "./context.jsx";
 
-export { data, RENDER, EpisodesContext, EpisodesProvider };
+export { data, RENDER, reducer, PREV_EPISODE, NEXT_EPISODE, SET_ACTIVE_INDEX, SET_OVERLAY_ACTIVE, EpisodesContext, EpisodesProvider };
```
pookage: continuing from above - you should be able to just export `ACTIONS` here.

src/components/Episodes/reducer.jsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import {NEXT_EPISODE, PREV_EPISODE, SET_OVERLAY_ACTIVE, SET_ACTIVE_INDEX} from "./";
```
pookage: `ACTIONS`

src/components/Episodes/reducer.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default function reducer(state, action){
+	const {
+		type, 
+		value
+	} = action;
+
+	switch(type){
+		case NEXT_EPISODE:
```
pookage: `ACTIONS.NEXT_EPISODE` etc etc (I won't duplicate for the others)

src/components/Episodes/reducer.jsx
```diff
@@ -0,0 +1,35 @@
...
+			return {
+				...state,
+				activeIndex: state.activeIndex-1
+			}
+		case SET_ACTIVE_INDEX:
+			return {
+				...state,
+				activeIndex: value
+			}
+		case SET_OVERLAY_ACTIVE:
```
pookage: see earlier comments re: naming actions

src/components/Episodes/reducer.jsx
```diff
@@ -0,0 +1,35 @@
...
+			return {
+				...state,
+				overlayActive: value
+			}
+		default:
+			throw new Error("Invalid action requested.");
+			break;
```
pookage: Nice - good shout, although three points!
- This will cause everything to fail and crash - if that's what you want then keep as is, if you just want to alert yourself, then `console.error()` is also available
- You don't need an instance of `Error` to throw something - you can just do a string unless you specifically want to use stuff from the `Error` class.
- Whenever you're throwing developer errors like this - make debugging as helpful as possible by giving yourself an indication of what went wrong - ie 

```
throw `Invalid action: ${type} requested with a value of ${value}`;
```

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -1,22 +1,21 @@
 import React, { useContext } from "react";
 import { 
 	data,
+	SET_ACTIVE_INDEX,
```
pookage: `ACTIONS`

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -1,22 +1,21 @@
...
-			activeIndex
+		state: {
+			activeIndex,
+			episodeCount
 		},
-		set: {
-			activeIndex: setActiveIndex
-		}
+		dispatch
 	} = useContext(Episodes);
```
pookage: `.state` as mentioned earlier

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -33,7 +32,10 @@ export default function EpisodesNav(){
 	function handleNavPressed(e, newIndex){
 		e.preventDefault();
 
-		setActiveIndex(newIndex);
+		dispatch({
+			type: SET_ACTIVE_INDEX,
```
pookage: descriptive action names, as above.

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -2,7 +2,8 @@ import React, { useContext } from "react";
 import { s } from "./";
 import Link from "COMPONENTS/Link/";
 import { 
-	EpisodesContext as Episodes 
+	EpisodesContext as Episodes,
+	SET_ACTIVE_INDEX
```
pookage: `ACTIONS`

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -11,12 +12,10 @@ function renderPaginationLink(data, index){
...
 	const { 
-		get: {
+		state: {
 			activeIndex 
 		},
-		set: {
-			activeIndex: setActiveIndex
-		}
+		dispatch
 	} = useContext(Episodes);
```
pookage: `.state`

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -33,7 +32,10 @@ function renderPaginationLink(data, index){
 	// CLICK HANDLER
 	// --------------------------
 	function handlePageClick(){
-		setActiveIndex(index);
+		dispatch({
+			type: SET_ACTIVE_INDEX,
```
pookage: descriptive actions as mentioned earlier.

src/shared/utils.js
```diff
@@ -1,6 +1,6 @@
 function convertToSafeString(string, seperator="_"){
-	return string.replace(/\W+/g, '') // remove non-alphanumeric
-			.replace(/ /g, seperator)      // replace spaces with underscores or provided seperator
+
+	return string.replace(/\W+/g, seperator) // remove non-alphanumeric & space, inset seperators
 			.toLowerCase();
 }//convertToSafeString
```
pookage: is this a duplicate of the change you made in the other branch? heads-up whenever you do something like that! If they're exactly the same then you should be able to merge them without any issues, but this is how things tend to fall out of sync and cause problems. If you want to expand on the work done from another branch you can always checkout a new branch _from that branch_ instead of `develop` (or whatever our base is)

fshenton: Not a duplicate, it was a bug fix to the change I made on the episode-preview branch. Having the two replace functions was doing something weird and resulting in strings with no spaces and no seperators.

src/components/Episode/Episode.jsx
```diff
@@ -8,9 +8,11 @@ export default function Episode(props){
...
 	//---------------------------
 	const {
-		episodeCount,
-		activeIndex
-	} = useContext(Episodes).get;
+		state: { 
+			activeIndex,
+			episodeCount
+		}
+	} = useContext(Episodes);
```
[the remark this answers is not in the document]
fshenton: Done

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -1,5 +1,7 @@
 import React, { useContext } from "react";
 import {
+	SET_ACTIVE_INDEX,
+	SET_OVERLAY_ACTIVE,
```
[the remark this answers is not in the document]
fshenton: Cool, done. That's what i wanted to do! :)

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -55,21 +54,26 @@ export default function EpisodePreview(props){
...
 
-
 	// CLICK HANDLERS
 	// -----------------------------
 	function handleNavPressed(e, newIndex){
 		e.preventDefault();
 
-		setActiveIndex(newIndex);
+		dispatch({
+			type: SET_ACTIVE_INDEX,
```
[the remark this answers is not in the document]
fshenton: Good shout.
Chose:
- GET_EPISODE as it's simple and (I think) clear enough (GET_EPISODE(1))
- SHOW_TRAILER for the same reasons. Hard to think of perfect verb but I like SHOW_TRAILER(true/false)

src/components/EpisodesNav/renderUtils.jsx
fshenton: `.state` wouldn't help here as we're destructuring dispatch also, which is state's sibling.

PR #21, opened 2019-09-11
Markup refactor

fshenton:
Mostly just minor spring cleaning.
CharactersContext now uses useState to control activeIndex and it's setActiveIndex function.
Prev/next (named) buttons now update the activeIndex and the aria-hidden values are correctly updating.

I'm sure there are some inconsistencies, but I feel that once the Episodes work is ready to come in, the Markup branch will be in a good state ahead of layout work.


src/components/Character/renderUtils.jsx
```diff
@@ -1,19 +1,47 @@
...
+
+	// CONTEXT
+	// ---------------------------------
+	const {
+		get: {
+			activeIndex
+		},
+		set: {
+			activeIndex: setActiveIndex
+		}
```
pookage: Ahhh, if this is what you were referring to as to the `useState` stuff? Yeah, for this we do want to do the `useReducer` stuff then.  Basically the rule is:

- if it's only changing the state in itself then do `useState`
- if it's changing other elements as well, then do `useContext` and manage that context with `useReducer` if stuff in it needs to be changed.

This is just to keep implementations the same throughout - otherwise you gotta stop and think about which state management thing to use and it's better to just have a solid rule to follow. At least it'll be easy if it's only the active index that needs changing!

src/components/Character/renderUtils.jsx
```diff
@@ -1,19 +1,47 @@
...
+
+	//RENDER
+	// ---------------------------------
 	const safeName       = UTILS.convertToSafeString(name);
 	const directionLabel = direction > 0 ? "Next" : "Previous";
 
+
+	// CLICK HANDLER
+	// ---------------------------------
+	function changeCharacter(e){
```
pookage: I'd probably name this `updateCharacterIndex`

src/components/Characters/context.jsx
```diff
@@ -20,11 +11,22 @@ function CharactersProvider(props){
...
 
+	const [ activeIndex, setActiveIndex ] = useState(initialState.activeIndex);
+
+	const state = {
+		get: {
+			activeIndex
+		},
+		set: {
+			activeIndex: setActiveIndex
+		}
```
pookage: as mentioned earlier - let's do the `useReducer` here.

PR #27, opened 2019-09-16
Layout about

fshenton:
Notes:
- using an App copy so I can isolate the About page easily. I figured that was cleaner than adding aria-hidden to all other components, which would trigger new aria-hidden styling (display: none) in global.scss.
- Didn't actually make any changes specifically for tablet or for desktop portrait, as they look "OK". 
     - Might be good to have the scrolling paragraphs I used in landscape desktop, but wanted to get your thoughts first.
- Created additional body text values (body1-4) to test the overflow with longer text. Figured having separate `<p>`s made more sense than one long string that contained forced linebreaks etc.
- Spacing between header/subheader (in particular) could be tighter (closer to design) but it's roughly there. Again, wanted to get your eyes on and see what you think before tweaking.


src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+			background-color: pink;
+
+			.heading {
+				margin-bottom: 1rem;
+				font-size: 1.15rem;
+			}
+
+			.subheading {
+				margin-bottom: 1.5rem;
+				font-size: calc(1.15rem * 1.62);
```
pookage: This is a magic number - we should indicate where it's coming from by using variables:

```
font-size: calc(var(--font_size__heading) * 1.62)
```

src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+			height: 100%;
+			padding-top: var(--landscape_spacing);
+			padding-right: var(--landscape_spacing);
+			padding-bottom: 0;
+			padding-left: var(--landscape_spacing);
+			overflow-y: auto;
+
+			.heading {
+				margin-bottom: 0.5rem;
+				font-size: calc(1rem * 0.62);
```
pookage: I mean...not sure if this is any more readable than `0.62rem`? 😬

src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			.subheading {
+				margin-bottom: 1rem;
+				font-size: 1rem;
+			}
+
+			.bodyContainer {
+				margin-bottom: 1rem;
+				font-size: 0.5rem;
```
pookage: the body `font-size` should _always_ be `1rem` or `1em` - everything else should scale around that, as that's the constant that everything else is based around.

src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+				height: 100%;
+				padding-top: 20vh;
+				overflow-y: hidden;
+				
+				.subheading {
+					font-size: calc(1.15rem * 1.62);
+				}
+
+				.bodyContainer {
+					height: 53vh;
```
pookage:
```suggestion
					height: 100%;
					max-height: 53vh;
```
We only want it to stop growing once it gets too big, we don't want to limit it more than we need to

src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+
+			.heading {
+				margin-bottom: 1rem;
+				font-size: 1.15rem;
+			}
+
+			.subheading {
+				margin-bottom: 1.5rem;
+				font-size: calc(1.15rem * 1.62);
+				line-height: 1.5em;
```
pookage: is this standard for subheadings? should this be in `fonts.scss`?

src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+	}
+}
+
+
+@media only screen and (min-width: 640px) and (max-width: 767px) {
+	.wrapper {
+		flex-direction: row;
+		--landscape_spacing: 5vw;
```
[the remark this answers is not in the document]
fshenton: Cool, done.

src/components/About/styles.scss
```diff
@@ -0,0 +1,125 @@
...
+			height: 38%;
+		}
+
+		.textContainer {
+			width: 100%;
+			height: 62%;
+			padding-top: var(--portrait-spacing);
+			padding-right: var(--portrait-spacing);
+			padding-bottom: 0;
+			padding-left: var(--portrait-spacing);
```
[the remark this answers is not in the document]
fshenton: The shorthand didn't work! 
Is there an issue with using variables in shorthand expressions like that?

fshenton:
@pookage 
Areas i'm still a big foggy on:
- best way to scale fonts as we move up the media break points
    - should we be defining different sizes for diff device types in our fonts?
    - my current (iffy) approach was using a _factor_ variable that scales the default font size variables
- not really sure what to do with desktop portrait

src/components/About/About.jsx
```diff
@@ -1,41 +1,44 @@
- import React from "react";
+import React from "react";
+import { renderBody as render } from "SHARED/renderUtils.jsx";
```
pookage: we should probably export an object called `RENDER` from the shared renderUtils, with `renderBody` in there, so that we can still do things like `RENDER.body(data)`

src/components/About/About.jsx
```diff
@@ -1,41 +1,44 @@
...
+		body
 	} = data;
 
 	const { 
 		src, 
 		description 
-	} = image;
+	} = heroImg;
+
+	const bodyContent = render(body); // create spacing at line-breaks
```
pookage: so, what I usually do to solve this naming problem, is alias the data-object as `**Data` to preserve the best variable names for the render return, so in this case we'd have:

```
const {
    ...,
    body: bodyData
} = data;

const body = RENDER.body(bodyData);
```

src/components/App/App.jsx
```diff
@@ -8,28 +8,29 @@ import Landing from "COMPONENTS/Landing/";
...
 import Characters from "COMPONENTS/Characters/";
 import Episodes from "COMPONENTS/Episodes/";
+import Fonts from "SHARED/fonts.scss";
+import { s } from "./";
 
 export default function App(){
 
 	return (
-		<div>
+		<div className={ s.container }>
```
pookage: this should be `.wrapper`

src/components/App/App.jsx
```diff
@@ -8,28 +8,29 @@ import Landing from "COMPONENTS/Landing/";
...
-			<Navigation />
+			{ false && <Navigation /> }
 
 			{/* PAGES */}
-			<main>
-				<Landing />
-				<About />
-				<Characters />
-				<Episodes />
+			<main className={ s.main }>
```
pookage: this should be `.container`

src/components/App/styles.scss
```diff
@@ -0,0 +1,3 @@
+.container, .main {
```
pookage: `.wrapper, .container` as per above comments (haven't done as a commit suggestion here in-case there were problems)

src/components/About/About.jsx
pookage: oop, OR, we could do

```
import * as RENDER from "SHARED/renderUtils.jsx";
```

although we'd have `RENDER.renderBody` instead of `RENDER.body` - I'll let you figure that fellah out!

src/components/About/styles.scss
```diff
@@ -0,0 +1,148 @@
...
+	.hero {
+		display: block;
+		width: 100%;
+		height: 38%;
+	}
+
+	.textContainer {
+		width: 100%;
+		height: 62%;
+		--padding__body: 10vw;
```
pookage: let's put all the variables as the very last thing in a class declaration - they look so different and weird that it feels odd to have them mixed in.

src/components/About/styles.scss
```diff
@@ -0,0 +1,148 @@
...
+// Tablet portrait
+@media (min-width: 768px) {
+	.hero {
+		height: 60%;
+	}
+
+	.textContainer {
+		height: 40%;
+		--font_size__factor: 1.62;
+		font-size: calc(var(--font_size__body)*var(--font_size__factor));
```
pookage: This is basically the equivalent of saying `1.62em`

src/components/About/styles.scss
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+		.heading {
+			font-size: calc(var(--font_size__heading)*var(--font_size__factor));
```
pookage: yerr, I can see why this felt weird - this is definitely where `em` comes in instead of `rem` - this is fine for now, but we should maybe take a look at this at some point together to find a neat solution for it.

src/components/About/styles.scss
```diff
@@ -0,0 +1,148 @@
...
+
+// Tablet landscape
+@media (min-width: 768px) and (orientation: landscape) {
+	.hero {
+		height: 100%;
+	}
+
+	.textContainer {
+		height: 100%;
+		--font_size__factor: 1.25;
```
pookage: aah, so that's why you did it - yeah, we can deffo use `em` cleverly here instead, and just be changing the font-size...might mean another wrapping `<div>` or something though. It's a clever use of the variables, but I think as we have `em` as a native way to scale fonts it feels like we should be using that instead. Grab me after work tomorrow and we'll find a way to do this.

src/components/About/About.jsx
fshenton: Using:
import RENDER from "SHARED/renderUtils.jsx"
&&
const body = RENDER.body(bodyData);

Quite close to Episodes approach, so seems good to me :)

fshenton:
I've added a quick and dirty font-size change, using rem and em. Feels better than using the calc and var stuff which was pretty messy to look at. Happy to revisit later if there's a better way of doing it, but it looks good enough for now!

PR #31, opened 2019-09-18
Feature/layout/navigation

fshenton:
Notes:
- Struggled again with getting usable font sizes from the designs. Ended up eye balling a lot of it and tweaking manually in inspector to find good values.
- The media breakpoint for desktop portrait doesn't work (at least for me) as my half screen is less than 1200px wide. Is that not a problem for you because you have a higher resolution set on your PC? Either way, what's the best way to emulate that?


pookage:
> * The media breakpoint for desktop portrait doesn't work (at least for me) as my half screen is less than 1200px wide. Is that not a problem for you because you have a higher resolution set on your PC? Either way, what's the best way to emulate that?

Not really, unfortunately, but that would mean that snapped-left would basically just be portrait tablet, and if that's fine then it's fine 👍

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Desktop landscape
+@media (min-width: 1200px) {
+	.wrapper {
+		font-size: 1.2rem;
+	}
+
+		.button {
+			top: 1.875rem;
```
pookage: following my above example on the whole `<NavToggle />` thing, for desktop, I get
```
.wrapper {
    top: 3.33em;
    .font-size: 1rem;
}

File: src/components/Navigation/styles.scss [hunk 2]
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		.menu {
+			padding-top: 20vh;
+		}
+
+			.link {
+				font-size: 4.5em;
```
pookage: Once you've re-done all the scaling maths etc, whatever the result is should go into the `.container` (`.menu` currently) rather than on the individual item.  CSS is all about the cascade - so you want to specify stuff as it makes sense to - specifying on the `.item` here would only make sense if the parent already had a different `font-size` and you were scaling off of that, if you get me?

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,133 @@
...
+
+			&:nth-child(1){
+				position: absolute;
+				right: 0;
+			}
+		}
+
+	.menu {
+		display: flex;
+		padding-top: 20vh;
```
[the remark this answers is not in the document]
fshenton: I avoided this because it wasn't dead in the middle vertically. Much easier this way though !

fshenton:
Thanks @pookage for all the feedback, much appreciated!
I followed what you suggested, and then did some tweaking here and there to get it looking closer to the designs. Looking pretty good now :)

src/components/Navigation/Navigation.jsx
```diff
@@ -42,22 +43,32 @@ export default function Navigation(){
...
+						menu
+					</span>
+					<span 
+						className={`${s.label} ${s.close}`}
+						aria-hidden="false"
+					>
+						close
+					</span>
 				</span>
 			</button>
```
pookage: Can we move this into its own `<NavToggle>` component pretty-please? 🙏  Just create a new `Navigation` context and have the new component dispatch a `OPEN_NAVIGATION` action or what-have you 😋

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,117 @@
...
+
+	.button {
+		display: inline-flex;
+		position: fixed;
+		top: 1.15em;
+		right: 1em;
+		align-items: center;
+		font-size: 1.5rem;
+		
+		&:before {
```
pookage:
```suggestion
		&::before {
```
pseudo-element!

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,117 @@
...
+			content: "";
+			display: block;
+			width: 1em;
+			height: 1.15em;
+			font-size: 1.33em;
+			background: url(images/icons/cross.svg) center / contain no-repeat;
```
pookage: not for this pass, but in the details pass this `::before` element will likely mutate into a `<span>` with its own `::before` and `::after` so that we can do the fancy cross animation thing - kinda another reason why it'll be good to have its own component; it'll have its own cleverness going on.

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+	
+// Mobile landscape
+@media (orientation: landscape) {
+	.button {
+		font-size: 1rem;
+	}
+
+	.container {
```
pookage: I don't think anything is using this any more!

src/components/Navigation/styles.scss
fshenton: Ah it worked with one : but looks like that's CSS2 and :: is CSS3.
Changed anyway so it's most up to date.

src/components/Navigation/Navigation.jsx
fshenton: OK, reducer added to navigation which the new NavToggle uses to change the open boolean.
Still using a state in Navigation as it needs to keep aria-expanded updated, it does this by passing a callback function to NavToggle.

Also did styling for Socials. Pretty basic but it's there 😄

src/components/NavToggle/NavToggle.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+	const {
+		state: {
+			open: isOpen
+		},
+		dispatch
+	} = useContext(Nav);
+
+	const {
+		setExpanded //callback for Navigation comp
```
pookage: You don't need this fellah! this is all handled with the dispatch

src/components/NavToggle/NavToggle.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+		dispatch({
+			type: ACTIONS.OPEN_NAVIGATION,
+			value: !isOpen
+		});
+	}// toggleOpen
+
+	return (
+			<button 
+				className={ s.button }
```
pookage:
```suggestion
				className={ s.wrapper}
```
outermost element - fellah's a wrapper

src/components/NavToggle/NavToggle.jsx
```diff
@@ -0,0 +1,58 @@
...
+				role="switch"
+				aria-checked={ isOpen }
+				aria-controls="navigation__links"
+				onClick={ toggleOpen }
+			>
+				<span className={ s.labelContainer }>
```
pookage:
```suggestion
				<span className={ s.container}>
```
no need prepend it with `label` now that its in its own component

src/components/Navigation/Navigation.jsx
```diff
@@ -37,32 +32,26 @@ export default function Navigation(){
...
-					Menu
-				</span>
-				<span>
-					Close
-				</span>
-			</button>
+		<nav className={ s.wrapper }>
+			<NavProvider>
+				<NavToggle 
+					setExpanded={ setExpanded }
```
pookage:
```suggestion
```
As a rule of thumb, now that you know about contexts, there's not a need to be passing props down like this anymore 👌

src/components/Navigation/Navigation.jsx
```diff
@@ -1,20 +1,15 @@
...
-	const [ open, setOpen ] = useState(false);
-
-
-	//EVENT HANDLING
-	//---------------------------
-	function toggleOpen(){
-		setOpen(!open);
-	}//toggleOpen
-
+	let [expanded, setExpanded] = useState(false);
```
pookage:
```suggestion
```
We should use the state from the `NavigationProvider here instead.

src/components/Navigation/Navigation.jsx
```diff
@@ -37,32 +32,26 @@ export default function Navigation(){
 		}
 	];
 
-
-	const links       = linkData.map(RENDER.item);
-	const ariaOpen    = open.toString();
+	const links      	  = linkData.map(RENDER.item);
+	const ariaExpanded    = expanded.toString();
```
pookage:
```suggestion
	const expanded    = expanded.toString();
```
once we've removed the state we can put this back to its non-namespaced variant

src/components/Navigation/Navigation.jsx
```diff
@@ -37,32 +32,26 @@ export default function Navigation(){
 		}
 	];
 
-
-	const links       = linkData.map(RENDER.item);
-	const ariaOpen    = open.toString();
+	const links      	  = linkData.map(RENDER.item);
```
pookage: Ooop, looks like there might be another tab in here...

src/components/Navigation/Navigation.jsx
```diff
@@ -37,32 +32,26 @@ export default function Navigation(){
...
-			>
-				<span>
-					Menu
-				</span>
-				<span>
-					Close
-				</span>
-			</button>
+		<nav className={ s.wrapper }>
+			<NavProvider>
```
pookage: if we take the provider out here and wrap this whole `Navigation` component within it (ie. in `<App />`, then we can use its `open` state instead of all the `expanded` stuff in here.

src/components/Navigation/Navigation.jsx
```diff
@@ -1,20 +1,15 @@
 import React, { useState } from "react";
 import Socials from "COMPONENTS/Socials/";
-import { RENDER } from "./";
+import NavToggle from "COMPONENTS/NavToggle";
+import { 
+	RENDER, 
+	NavProvider 
+} from "./";
+import { s } from "./";
```
pookage:
```suggestion
```
We're importing from here in the line above, so this should come from here too.

src/components/Navigation/Navigation.jsx
```diff
@@ -1,20 +1,15 @@
 import React, { useState } from "react";
 import Socials from "COMPONENTS/Socials/";
-import { RENDER } from "./";
+import NavToggle from "COMPONENTS/NavToggle";
+import { 
+	RENDER, 
+	NavProvider
```
pookage:
```suggestion
	NavProvider,
	s
```

src/components/NavToggle/styles.scss
```diff
@@ -0,0 +1,78 @@
...
+				&[aria-hidden="false"]{
+					opacity: 1;
+				}
+
+				&.open {
+					position: absolute;
+					top: 0;
+					right: 0;
+				}
+			}
```
pookage: oop, looks like in the copy+paste you've got one tab too many? need to bring it in one step.

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
+// Mobile portrait
+.wrapper {
+	width: 100%;
+	height: 100%;
+	padding-top: 10vh;
```
pookage:
```suggestion
```
This seems like a magic number that's going to mess with our vertical centering.

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+	background-color: red;
+}
+
+	.itemContainer {
+		display: flex;
+		width: 100%;
+		height: 75vh;
```
pookage:
```suggestion
		height: 75%;
```
Try not to use solid units like `vh`, `px` etc when you've used relative units in the parent - we've already set ourselves up to use `%` in the above, and if we use `vh` down here then it has the opportunity to break - for example, we want the component to adapt if the `.wrapper` is set to `height: 30vh` for example - we'd want everything inside it to adjust to accomodate that, whereas this would break it.

src/components/Navigation/Navigation.jsx
```diff
@@ -37,32 +32,26 @@ export default function Navigation(){
...
 			<ul 
 				id="navigation__links"
-				aria-expanded={ ariaOpen }
+				className={ s.itemContainer }
+				aria-expanded={ ariaExpanded }
 			>
 				{ links }
 			</ul>
-			<Socials />
+			<div className={ s.socials }>
```
pookage: Can we lose this wrapper? I don't think we need it.

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
+// Mobile portrait
+.wrapper {
+	width: 100%;
+	height: 100%;
+	padding-top: 10vh;
+	background-color: red;
+}
+
+	.itemContainer {
```
pookage:
```suggestion
	.container {
```
no conflict with the nav button anymore 🙌

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+				margin-top: 0.7em
+			}
+
+			.link {
+				text-decoration: none;
+			}
+		}
+
+	.socials {
+		font-size: 1rem;
```
pookage: This should just be specified in the component itself

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+
+		li+li {
+			margin-left: 4em;
```
pookage: we can define this in the social styles

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+	}
+	
+// Mobile landscape
+@media (orientation: landscape) {
+	.itemContainer {
+		margin-bottom: 0.5em;
+		font-size: 3rem;
+	}
+
+	.socials {
```
pookage: don't forget to remove these

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
+// Mobile portrait
 .wrapper {
+	display: flex;
+	justify-content: center;
+	width: 100%;
 	font-size: 1em; //nothing comment to make sure class appears
```
pookage:
```suggestion
	font-size: 1rem;
```

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+	
+		li {
```
pookage:
```suggestion
		.item {
```
we can give it this in the `renderUtils` - in `renderItem` I think.

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+
+}
+
+// Desktop landscape
+@media (min-width: 1200px) {
+	.itemContainer {
+		margin-bottom: 0;
+	}
+
+	.socials {
```
pookage: don't forget to remove these fellahs

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+			width: 1em;
+			height: 1em;
+			list-style: none;
```
pookage:
```suggestion
			list-style: none;
			&+.item {
				margin-left: 4em;
			}
```
Implementing the rule you'd established in the `Navigation` styles here instead of up there.

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+		}
+
+			a {
+				display: block;
+				width: 100%;
+				height: 100%;
+				background-color: blue;
+
+				&[aria-label="Facebook"]{
+					background-image: url("path/to/facebook.jpg");
```
pookage:
```suggestion
					background-image: url("path/to/facebook.svg");
```
Just me being picky - we know that we need transparency so it'll either be a `svg` or a `png`, and it's a nice simple shape so we'll probably go for `svg`.  All of the 404s that appear in the console will basically become our shopping list of assets in the details stage.

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+				}
+
+				&[aria-label="Twitter"]{
+					background-image: url("path/to/twitter.jpg");
```
pookage:
```suggestion
					background-image: url("path/to/twitter.svg");
```

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+				}
+
+				&[aria-label="Instagram"]{
+					background-image: url("path/to/instagram.jpg");
```
pookage:
```suggestion
					background-image: url("path/to/instagram.svg");
```

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+	width: 100%;
 	font-size: 1em; //nothing comment to make sure class appears
+	
+		li {
+			width: 1em;
+			height: 1em;
+			list-style: none;
+		}
+
+			a {
```
pookage:
```suggestion
			.link {
```
We should give this a name in the `renderUtil`

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+				display: block;
+				width: 100%;
+				height: 100%;
+				background-color: blue;
+
+				&[aria-label="Facebook"]{
```
pookage:
```suggestion
				&.facebook{
```
Neat idea, but this would break if, for example, we find that "Visit us on Facebook" works better as a label for engagement or SEO or whatever; this fits under the category of 'content', I reckon, and we want to avoid tying styling to content where we can.

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+					background-image: url("path/to/facebook.jpg");
+				}
+
+				&[aria-label="Twitter"]{
```
pookage:
```suggestion
				&.twitter{
```

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,64 @@
...
+					background-image: url("path/to/twitter.jpg");
+				}
+
+				&[aria-label="Instagram"]{
```
pookage:
```suggestion
				&.instagram{
```

src/components/Navigation/Navigation.jsx
fshenton: Cool, wasn't sure the best way to do this :)

src/components/Socials/styles.scss
fshenton: Need some advice on how to do className={ s.<insertsocialname>}. Sent you a message separately

src/components/Navigation/styles.scss
fshenton: Removed and changed height of container to 100%. Seems to be aligning fine now.

src/components/App/App.jsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import HomePermalink from "COMPONENTS/HomePermalink/";
 import HistoryChannel from "COMPONENTS/HistoryChannel/";
 import WatchNow from "COMPONENTS/WatchNow/";
-import Navigation from "COMPONENTS/Navigation/";
+import Navigation, { NavProvider } from "COMPONENTS/Navigation/";
```
pookage: did we want to do `{ Provider as NavProvider }` here? Trying to think about the best habits to form, and that seems like a good one...

src/components/NavToggle/NavToggle.jsx
```diff
@@ -0,0 +1,55 @@
...
+				className={ s.wrapper }
+				role="switch"
+				aria-checked={ isOpen }
+				aria-controls="navigation__links"
+				onClick={ toggleOpen }
+			>
+				<span className={ s.container }>
+					<span 
+						className={`${s.label} ${s.open}`}
+						aria-hidden="true"
```
pookage:
```suggestion
						aria-hidden={isOpen}
```

src/components/NavToggle/NavToggle.jsx
```diff
@@ -0,0 +1,55 @@
...
+					>
+						menu
+					</span>
+					<span 
+						className={`${s.label} ${s.close}`}
+						aria-hidden="false"
```
pookage:
```suggestion
						aria-hidden={!isOpen}
```

src/components/Navigation/Navigation.jsx
```diff
@@ -37,32 +37,22 @@ export default function Navigation(){
 		}
 	];
 
-
-	const links       = linkData.map(RENDER.item);
-	const ariaOpen    = open.toString();
+	const links      = linkData.map(RENDER.item);
+	const expanded   = open.toString();
```
pookage: what conclusion did we reach with `toString()` was it needed in the end, or was it just with vanilla javascript that you need to stringify it?

src/components/Navigation/Navigation.jsx
pookage: Oop, it's still here 😬 - if there's a reason that `<Socials />` needs to be wrapped then let's totally discuss that when we go over stuff later.

src/components/Socials/renderUtils.jsx
```diff
@@ -13,8 +14,12 @@ function renderItem(data){
 	const safeLabel = UTILS.convertToSafeString(label);
 
 	return (
-		<li key={ `social__${safeLabel}` }>
+		<li 
+			key={ `social__${safeLabel}` }
+			className={ s.item }
+		>
 			<Link
+				className={ `${s.link} ${s[safeLabel]}` }
```
pookage: aah - clever usage of `safeLabel` there! nice one.

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,70 @@
...
+			// &+.item {
+			// 	margin-left: 4em;
+			// }
+		}
+
+			.link {
+				display: block;
+				width: 100%;
+				height: 100%;
+				background-color: blue;
```
pookage:
```suggestion
				background-color: blue;
				background-size: contain;
				background-position: center;
				background-repeat: no-repeat;
```

src/components/Navigation/styles.scss
pookage: **poke** this would presumably in the `.wrapper` of `Socials/styles.scss`

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Tablet portrait
+@media (min-width: 768px) {
+	.container {
+		font-size: 5rem;
+	}
+
+	.socials {
+		width: 35%;
```
pookage: bump

src/components/Navigation/styles.scss
pookage: bump

src/components/Navigation/styles.scss
```diff
@@ -0,0 +1,87 @@
...
+				margin-top: 0.7em
+			}
+
+			.link {
+				text-decoration: none;
+			}
+		}
+
+	.socials {
+		width: 50%;
```
pookage: bump

src/components/NavToggle/styles.scss
```diff
@@ -0,0 +1,78 @@
...
+		font-size: 1rem;
+	}
+}
+
+// Tablet portrait
+@media (min-width: 768px) {
+	.wrapper {
+		right: 50%;
+		font-size: 2rem;
+		transform: translateX(50%);
```
pookage: This works, but I would default to
```
left: 50%;
transform: translateX(-50%);
```
as your go-to solution to this kinda thing, as the convention is to think from left -> right, but yeah - reality is that it works either way.

src/components/Navigation/Navigation.jsx
fshenton: Yarp, works fine without toString :)

src/components/App/App.jsx
fshenton: If you feel this is worth doing, I can do it but i'd there would be a fair bit of renaming as it is not how we're doing it elsewhere (for contexts and providers). Seems a bit minor to commit time to it

src/components/Navigation/Navigation.jsx
```diff
@@ -37,22 +37,19 @@ export default function Navigation(){
...
 				className={ s.container }
-				aria-expanded={ expanded }
+				aria-expanded={ open }
 			>
 				{ links }
 			</ul>
-			<div className={ s.socials }>
-				<Socials />
-			</div>
+			<Socials navOpen={ open } />
```
pookage: For this, I was thinking something more like `<Socials className={open ? s.open : s.closed} />`
If we wanted the socials to have access to `open` then we'd probably want to do `useContext` inside that component rather than pass it down as a props.

fshenton: Socials now using context as that seemed the nicest way to do it

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,83 @@
...
+	display: flex;
+	justify-content: space-between;
+	font-size: 2rem;
+	
+	&.closed {
+		opacity: 0;
+	}
+
+	&.open {
+		width: 50%;
```
pookage:
```suggestion
		width: 50%;
		max-width: 10em;
```
Otherwise on desktop they grow to a really large space that isn't similar to the designs.  I'm not sure if it's `10em`, but worth giving a quick measure.

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,83 @@
...
+
+				&.instagram {
+					background-image: url("path/to/instagram.svg");
+				}
+			}
 }
+
+// Mobile landscape
+@media (orientation: landscape) {
+	.wrapper.navOpen {
```
pookage:
```suggestion
	.wrapper.open {
```
aah, I think these are failing because it's `open` instead of `navOpen` now - don't forget to test after you make changes! 🙌

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,83 @@
...
+		width: 25%;
+	}
+}
+
+// Tablet portrait
+@media (min-width: 768px) {
+	.wrapper.navOpen {
```
pookage:
```suggestion
	.wrapper.open{
```

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,83 @@
...
+		width: 35%;
+	}
+}
+
+// Tablet landscape
+@media (min-width: 768px) and (orientation: landscape) {
+	.wrapper.navOpen {
```
pookage:
```suggestion
	.wrapper.open{
```

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,83 @@
...
+}
+
+// Desktop portrait
+@media (min-width: 1200px) and (max-aspect-ratio: 8/9) {
+	
+}
+
+// Desktop landscape
+@media (min-width: 1200px) {
+	.wrapper.navOpen {
```
pookage:
```suggestion
	.wrapper.open{
```

src/components/App/App.jsx
fshenton: Going to leave this for now.

src/components/Socials/Socials.jsx
```diff
@@ -1,25 +1,21 @@
-import React from "react";
-import { RENDER, s } from "./";
+import React, { useContext } from "react";
+import { RENDER, s, data } from "./";
+import { NavContext as Nav } from "COMPONENTS/Navigation/";
 
-export default function Socials(){
+export default function Socials(props){
```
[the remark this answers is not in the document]
fshenton: I passed props this way in the new BackgroundVideo component, works fine!

src/components/Socials/styles.scss
fshenton: Went with 6em, looks better. Not sure how I missed the navOpen class name still being used 🙈

src/components/Socials/styles.scss
```diff
@@ -1,3 +1,52 @@
...
+			}
 }
+
+
+
+
+
+
+
+
```
pookage: lot of empty lines here, but no biggie.

src/components/NavToggle/NavToggle.jsx
```diff
@@ -0,0 +1,55 @@
...
+					</span>
+					<span 
+						className={`${s.label} ${s.close}`}
+						aria-hidden={ !isOpen }
+					>
+						close
+					</span>
+				</span>
+			</button>
+	)
```
pookage: We might have one indentation too many here?

PR #36, opened 2019-09-19
Feature/layout/landing

fshenton:
Notes:
- This went pretty smoothly, apart from the usual unreliable font sizing and spacing from design.
- I'm ready, but not necessarily excited, about you going into a frothing rage when you see I used 0.7em for the body text. Pls don't desecrate my body once I'm dead.

Tdlr: results are decent/good, but the manual tweaking of values gained from designs feelsbadman.


src/components/Landing/styles.scss
```diff
@@ -0,0 +1,75 @@
...
+	font-size: 1.25rem;
+	line-height: 1;
+}
+
+	.heading, .subheading {
+		font-family: var(--font_family__heading);
+	}
+
+	.heading {
+		margin-bottom: 0.1em;
```
pookage:
```suggestion
		margin-bottom: 0.6em;
```
I'm getting `36px` here, which would be `0.6em` (also 2em of the body, which is neat)

src/components/Landing/styles.scss
```diff
@@ -0,0 +1,75 @@
...
+		font-size: 2.9em;
+		letter-spacing: 0.1em;
+		text-transform: uppercase;
+	}
+
+	.subheading {
+		margin-bottom: 0.5em;
```
pookage:
```suggestion
		margin-bottom: 1em;
```
I'm getting `45px` here, so a nice, round `1em` from me.

src/components/Landing/Landing.jsx
```diff
@@ -1,18 +1,19 @@
...
+		<header className={ s.wrapper }>
+			<h1 className={ s.heading }>
 				Vikings
 			</h1>
-			<h2>
+			<h2 className={ s.subheading }>
 				Descend into Darkness
 			</h2>
-			<p>
+			<p className={ s.body }>
```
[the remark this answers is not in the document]
fshenton: K this works, apart from it's including the spans as part of the string. What's a good workaround for this?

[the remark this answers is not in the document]
fshenton: ALSO, I want to pass the component name to the render util but it's not working the map function won't accept additional arguments. It uses the full text as the key prefix, as it's expecting an array as the third argument 🤣

src/components/Landing/Landing.jsx
```diff
@@ -1,18 +1,19 @@
...
 				When a king falls, another will rise. But who will it be?
 			</p>
-			<p>
+			<p className={ s.body }>
 				Get a closer look at <span>Vikings season 5</span>.
```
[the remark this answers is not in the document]
fshenton: Pulled the span content out and added as an `<a>` instead to the return. For some reason the remaining string doesn't include a space, so we'd have to add some in. Bit weird

src/components/Landing/styles.scss
```diff
@@ -0,0 +1,75 @@
+// Mobile portrait
+.wrapper {
+	display: flex;
+	width: 100%;
+	height: 100%;
+	flex-direction: column;
+	justify-content: center;
+	align-items: center;
+	background-color: #FF60D7;
+	font-size: 1.25rem;
```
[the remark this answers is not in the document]
fshenton: I changed the rem value in one of the first layouts. Probably during tweaking phase (d'oh). So i've been assuming 1rem is 16 and working out em based on that.

src/shared/renderUtils.jsx
```diff
@@ -1,12 +1,12 @@
...
 
-function renderBody(body){
+function renderBody(body, component){
 	const paragraphs = body.split("\n"); // splits the string at every line-break
 	return paragraphs.map(renderParagraph);
 }// renderBody
 
-function renderParagraph(text, index){
-	const key = `about__body__paragraph_${index}`;
+function renderParagraph(text, index, component){
```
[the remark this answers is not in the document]
pookage: mmm, maybe name the variable `scope` or something too? Cos we may end up having multiple bodies per component - eg. for the gallery or character pages, so we'll need something other than the component name to prefix.

src/components/Landing/Landing.jsx
```diff
@@ -1,20 +1,30 @@
 import React from "react";
+import RENDER from "SHARED/renderUtils.jsx";
+import { s } from "./";
 
 export default function Landing(){
+
+	const bodyData = `When a king falls, another will rise. But who will it be?\nGet a closer look at `;
```
pookage: The 
```suggestion
	const bodyData = `When a king falls, another will rise. But who will it be?
	Get a closer look at `;
```
If we use a template literal then we don't have to use `\n` 🙌

pookage: Also, rather than split out the link like this, let's include it as part of the text - remember that we want to move all of the content like this to a `data.js` that we import, and that should all be treated as if it's client-facing via CMS or what-have-you.  We could do something like:

```
const data = {
    heading: Vikings,
    subheading: Desecend into Darkness,
    body: `When a king falls, another will rise. But who will it be?
Get a closer look at [Vikings Season 5](#about)`
};

export default data;
```

By using markdown we can put links in our data and then parse them out later - you can make your own `parseMarkdown` util if you'd like to - if not, a good tool for this is [marked](https://www.npmjs.com/package/marked) - which we can install and then import here.

src/components/Landing/Landing.jsx
```diff
@@ -1,20 +1,30 @@
...
 			</h2>
-			<p>
-				When a king falls, another will rise. But who will it be?
-			</p>
-			<p>
-				Get a closer look at <span>Vikings season 5</span>.
-			</p>
+			<div className={ s.body }>
+				{ body }
+				<a
```
pookage: Another reason we'd want to take the markdown appraoch is that 'Vkings season 5' is part of the sentence in the paragraph above, so we'd want it to be a child, not a sibling

src/shared/renderUtils.jsx
```diff
@@ -1,12 +1,12 @@
...
+	const key = `${component}__body__paragraph_${index}`;
 	return (
 		<p key={ key }>
```
pookage:
```suggestion
		<p className={s.paragraph} key={ key }>
```

src/components/Landing/styles.scss
```diff
@@ -0,0 +1,77 @@
...
+		margin-bottom: 1em;
+		font-size: 2.5em;
+	}
+
+	.body {
+		text-align: center;
+		font-size: 1em;
+		font-family: var(--font_family__body);
+
+		p+p {
```
pookage:
```suggestion
		.paragraph + .paragraph {
```
Just gotta update `renderUtils`

src/components/Landing/styles.scss
```diff
@@ -0,0 +1,77 @@
...
+			margin-top: 1em;
+		}
+
+		p:nth-child(2) {
+			display: inline-block;
+		}
+	}
+
+	.aboutLink {
```
pookage: don't forget to tidy-up these after

src/components/Landing/styles.scss
```diff
@@ -0,0 +1,77 @@
...
+	.body {
+		text-align: center;
+		font-size: 1em;
+		font-family: var(--font_family__body);
+
+		p+p {
+			margin-top: 1em;
+		}
+
+		p:nth-child(2) {
```
pookage: as mentioned earlier - we can lose this one once the anchor is back inline again.

src/shared/renderUtils.jsx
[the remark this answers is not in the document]
fshenton: cool, scope is the better word here for sure. Felt weird using `component`.
Updated and now binding the function to pass in scope. Looks to be working fine :)

src/shared/renderUtils.jsx
fshenton: As render util doesn't have own s.paragraph, now passing the class name in along with the scope.

src/components/Landing/Landing.jsx
fshenton: I did it that ^ way because I couldn't figure out a good way of parsing it and pulling the anchor out (i spent quite a bit of time on it!). 

I gave `marked` a try but after about 10 mins of head-scratching I decided to do it myself.

New findLink method in the render utils. I'm sure there's a nicer way of doing it but it was good to work it out and using a few things i've picked up :) (selective rendering woo!)

fshenton: @pookage

src/components/Landing/Landing.jsx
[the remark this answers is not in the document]
fshenton: Ended up using binding, interested to see if that rustles your jimmies @pookage 😋

src/components/Landing/Landing.jsx
```diff
@@ -1,20 +1,28 @@
...
 
 export default function Landing(){
+
+	const {
+		heading,
+		subheading,
+		body: bodyData
+	} = data;
+
+	const body = RENDER.body(bodyData, "landing", s.paragraph); //scope name for key, classname
```
pookage: now that stuff's getting more complication with this function, rather than have
```
renderBody(data, scope, className){ ... }
```
as our definition, can we instead do:
```
renderBody(data, options){
    const { scope, className } = options;
    ...
}
```
That way  we can use it like:
```
const body = RENDER.body(bodyData, { scope: "landing", className: s.paragraph });
```
or
```
const body = RENDER.body(bodyData, { className: s.paragraph });
```
and we don't need to worry about passing in `null` or `undefined` when we want to skip parameters.  This kinda format is used to distinguish between essential arguments and optional ones (with the optional ones existing in the `{}`)

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+	if(linkLabelStart === -1){
+		return false;
+	}
+	else {
+		//pull out the link label, discarding the braces
+		const linkLabelEnd = text.indexOf("]");
+		const linkText = text.substring(linkLabelStart+1, linkLabelEnd);
+
+		//pull out the link destination, discarding the braces
+		const linkDestStart = text.indexOf("(");
```
pookage:
```suggestion
		const destStart = labelEnd+1;
```
This can just be `labelEnd+1` as it will always follow the `]`

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
 
+function findLink(text){
+	const linkLabelStart = text.indexOf("[");
+
+	if(linkLabelStart === -1){
+		return false;
+	}
+	else {
+		//pull out the link label, discarding the braces
+		const linkLabelEnd = text.indexOf("]");
```
pookage:
```suggestion
		const labelEnd = markdown.indexOf("]", labelStart);
```
So, `indexOf` takes two arguments - string to find, and starting index; with the latter defautling to `0`.  So, what we can do here is use `linkLabelStart` to say where to start looking.

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+				<a href={ link.destination } >
+					{ link.text }
+				</a>
+			)}
 		</p>
 	);
 }// renderParagraph
 
+function findLink(text){
+	const linkLabelStart = text.indexOf("[");
```
pookage:
```suggestion
	const labelStart = markdown.indexOf("[");
```
we're already in a `findLink` method - probably don't need to namespace these 👌

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+
+	if(linkLabelStart === -1){
```
pookage:
```suggestion
	if(labelStart === -1){
```

pookage: Also, I'm going to leave this in here as it's personal preference, but you gotta choose here what format you're going to do throughout.  Either:
```
if(failCondition) return;

...do stuff...
```
or
```
if(passCondition){
    ...do stuff...
} else return;
```

The first one is called an 'early return' or an 'escape return' and it tends to be used by peeps who dislike indentation and so by returning early they don't need an `else`, and so the main body of the function can stay at level 1 indentation; whereas the latter is better for A/B comparissons and code-folding.

Either works, but this one here is a mishmash of both - an early return _and_ an `else` statement.  But yeah - whichever you choose, make sure you're consistant with it throughout your codebase.

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+		return false;
+	}
+	else {
+		//pull out the link label, discarding the braces
+		const linkLabelEnd = text.indexOf("]");
+		const linkText = text.substring(linkLabelStart+1, linkLabelEnd);
+
+		//pull out the link destination, discarding the braces
+		const linkDestStart = text.indexOf("(");
+		const linkDestEnd = text.indexOf(")");
```
pookage:
```suggestion
		const destEnd = markdown.indexOf(")", destStart);
```
As with `labelStart`, we can use `destStart` to say where to begin looking

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+		const linkDest = text.substring(linkDestStart+1, linkDestEnd);
```
pookage:
```suggestion
		const destination = markdown.substring(destStart+1, destEnd);
```

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+
+		return {
+			index: linkLabelStart,
+			text: linkText,
+			destination: linkDest
```
pookage:
```suggestion
			destination
```

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+function findLink(text){
+	const linkLabelStart = text.indexOf("[");
+
+	if(linkLabelStart === -1){
+		return false;
+	}
+	else {
+		//pull out the link label, discarding the braces
+		const linkLabelEnd = text.indexOf("]");
+		const linkText = text.substring(linkLabelStart+1, linkLabelEnd);
```
pookage:
```suggestion
		const text = markdown.substring(linkLabelStart+1, linkLabelEnd);
```

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+			{ link && (
+				<a href={ link.destination } >
+					{ link.text }
+				</a>
+			)}
 		</p>
 	);
 }// renderParagraph
 
+function findLink(text){
```
pookage:
```suggestion
function findLink(markdown){
```
let's specify what kind of text we're going to be parsing here.

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+		const linkText = text.substring(linkLabelStart+1, linkLabelEnd);
+
+		//pull out the link destination, discarding the braces
+		const linkDestStart = text.indexOf("(");
+		const linkDestEnd = text.indexOf(")");
+		const linkDest = text.substring(linkDestStart+1, linkDestEnd);
+
+		return {
+			index: linkLabelStart,
+			text: linkText,
```
pookage:
```suggestion
			text,
```

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+		const linkLabelEnd = text.indexOf("]");
+		const linkText = text.substring(linkLabelStart+1, linkLabelEnd);
+
+		//pull out the link destination, discarding the braces
+		const linkDestStart = text.indexOf("(");
+		const linkDestEnd = text.indexOf(")");
+		const linkDest = text.substring(linkDestStart+1, linkDestEnd);
+
+		return {
+			index: linkLabelStart,
```
pookage:
```suggestion
			index: labelStart,
```

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
-	return paragraphs.map(renderParagraph);
+
+	return paragraphs.map(
+		renderParagraph.bind(this, scope, className)
+	);
 }// renderBody
 
-function renderParagraph(text, index){
-	const key = `about__body__paragraph_${index}`;
+function renderParagraph(scope, className, text, index){
```
pookage: as with my comments on `renderBody`, let's create an `options` object for this, otherwise it's gonna get out of hand.

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,60 @@
...
+	const key = `${scope}__body__paragraph_${index}`;
+	
+	//if there is a link inside the text, cut it out
+	const link = findLink(text) || {};
+	if(link) text = text.slice(0, link.index);
```
pookage: This is neat, but it's gonna trip up when there are multiple links, so one to be aware of moving forward; although we don't need to address it just yet.

src/shared/renderUtils.jsx
fshenton: Cool, ty for the feedback on this.

I like both but prefer the simplicity of the early return. Updated.

pookage:
Before I do the PR, I'm thinking that maybe we should also create a `<LogoArtwork />` component at this stage, that at the very least has a `.wrapper` determines the height/width of everything - that way we can begin the details pass with a simple export of the V, and when it comes to doing all the really fancy shit, we already have a workspace to operate in.

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,59 @@
 import React from "react";
 
-function renderBody(body){
-	const paragraphs = body.split("\n"); // splits the string at every line-break
-	return paragraphs.map(renderParagraph);
+function renderBody(data, options = {}){ 
+	const paragraphs = data.split("\n"); // splits the string at every line-break
+
+	return paragraphs.map(
+		renderParagraph.bind(this, options)
```
pookage: quick couple notes here!

1. When we do a multi-line return, we should wrap inside of parenthesis, so this would be:

```
return (
    paragraphs.map(
        renderParagraph.bind(this, options)
    )
);
```

But also, it might be useful to give that bound `renderParagraph` a new name and keep it all single-line rather than multi-line it?  Something like:

```
const paragraphs = data.split("\n");
const customisedRenderParagraph = renderParagraph.bind(this, options);

return paragraphs.map(customisedRenderParagraph);
```

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,59 @@
...
+			)}
 		</p>
 	);
 }// renderParagraph
 
+
+function findLink(markdown){
+	const labelStart = markdown.indexOf("[");
+
+	if(labelStart === -1) return false; //no link found
```
pookage: I've created a monster 😂 😭

fshenton:
Added BackgroundVideo and LogoArtwork components. Sized roughly but seem OK on most devices.
Tried to use aspect ratio for Logo, not sure if it's working as well as we'd want!

src/components/BackgroundVideo/BackgroundVideo.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { s } from "./";
+
+export default function BackgroundVideo(props){
+
+	const {
+		videoSrc,
+		posterSrc = {}
```
pookage: It's inside of an attribute called `video`, so `props.video.videoSrc` seems uneccesary? I reckon something like

```
const props = {
    sources: {
        mp4: "path/to/video.mp4",
        webm: "path/to/video.webm"
    },
    poster: "path/to/poster.jpg"
};
```

src/components/BackgroundVideo/BackgroundVideo.jsx
```diff
@@ -0,0 +1,21 @@
...
+	} = props.video;
+
+	return (
+		<div className={ s.wrapper }>	
+			<video poster={ posterSrc }>
```
pookage: Before the `mp4`, if we include
```
<source
    src={ sources["webm"] }
    type="video/webm"
/>
```
Then we can save the user a hefty amount of filesize if they're using a browser thats supports webm! If it's not supported then it'll just fall-through and use the mp4 instead.

src/components/Landing/Landing.jsx
```diff
@@ -1,20 +1,38 @@
...
-			<p>
-				When a king falls, another will rise. But who will it be?
-			</p>
-			<p>
-				Get a closer look at <span>Vikings season 5</span>.
-			</p>
-		</header>
+		<>
+			<LogoArtwork />
+			<BackgroundVideo video={ video }/>
```
pookage: Needs updating to be `sources` and include a `poster` prop.

src/components/Landing/data.js
```diff
@@ -0,0 +1,12 @@
+const data = {
+	heading: "Vikings",
+	subheading: "Descend into Darkness",
+	body: `When a king falls, another will rise. But who will it be? 
+	Get a closer look at [Vikings Season 5](#about)`,
+	video: {
+		videoSrc: "path/to/video.mp4",
+		posterSrc: "path/to/poster.jpg"
```
pookage: as mentioned earlier - let's tweak this to `sources` and `poster`

src/components/Landing/styles.scss
```diff
@@ -0,0 +1,71 @@
...
+
+// Desktop portrait
+@media (min-width: 1200px) and (max-aspect-ratio: 8/9) {
+
+}
+
+// Desktop landscape
+@media (min-width: 1200px) {
+
+}
```
pookage: unuuuuused media queries!

src/components/LogoArtwork/styles.scss
```diff
@@ -0,0 +1,52 @@
+// Mobile Portrait
+.wrapper {
+	display: flex;
+	position: absolute;
+	height: 100%;
+	width: 100%;
```
pookage: Good use of the aspect-ratio stuff below, but I think that the wrapper should have the height/width defined down there - and the children should just be constrained to its own size; otherwise the implication is that the Logo Artwork fills the screen, when in-fact it doesn't.   To center it, we can just do our usual `transform` trick...

src/components/LogoArtwork/LogoArtwork.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { s } from "./";
+
+export default function LogoArtwork(){
+
+	return (
+		<div className={ s.wrapper }>
+			<img 
+				className={ s.logo }
+				src="path/to/file.png" alt=""
```
pookage:
```suggestion
				src="path/to/file.png" alt="Vikings Logo; an ornate 'V' carved out of wood." 
```

src/components/LogoArtwork/styles.scss
fshenton: Cool done, thanks!

PR #40, opened 2019-09-20
Feature/layout/character

fshenton:
Took out tweaked values, and used the ones i got from designs instead.
Interested to know your feedback bubsy :)
Wasn't sure if there was much/anything I should be doing with the image at this point, feels a bit pointless without the actual assets!


src/components/Character/styles.scss
```diff
@@ -1,19 +1,134 @@
...
+				}
+			}
+
+// Mobile portrait
+
+
+// Mobile landscape
+@media (orientation: landscape) {
+	.container {
+		top: 25vh;
```
pookage:
```suggestion
```
As mentioned above - we just need to center with flexbox here - don't need to push down from the top at all - but if we did, then we should be using `%` here - `vw` and `vh` are for when there's literally no other option. 

They're the weapons that come out, like, 3/4 of the way through the movie when your best friend gets killed and your return back to base, clothes torn, covered in blood; heading to the back of the cave where a slightly out-of-place bookshelf sits.  You shake your head in dismay as you reach out and pull out a book called _Absolutism_; causing the unit to slowly flip over revealing your secret, most powerful stash.

"I can't believe it's come to this", you say as you pick up a `vh` in one hand and a `vw` in the other. "But I have no other option".
"But...but...but wait, mr frank!"
You turn. The ghost of your now deceased amigo, `%`, is standin there, apalled.
"I've got to, little cento." You say, turning away. "There's a container out there whose height needs to be based on the viewport width..."

....

"n..n...no." 😭  little cento whimpers. "what has he become?"
🕶 🌆 🔫

src/components/Character/styles.scss
```diff
@@ -1,19 +1,134 @@
 .wrapper {
-	font-size: 1em;
+	position: absolute;
+	top: 0;
+	width: 100%;
+	height: 100%;
 
 	&[aria-hidden="true"]{
 		opacity: 0;
```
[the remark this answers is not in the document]
fshenton: Good shout! <3

src/components/Character/styles.scss
fshenton: 😭 poor lil cento!

fshenton: Poor lil cento 😭

src/components/Character/styles.scss
```diff
@@ -1,19 +1,134 @@
...
+			}
+
+			.previous {
+				left: var(--padding_container);
+
+				&:before {
+					display: inline-block;
+					content: "";
+					width: 1.36em;
+					height: 2em;
```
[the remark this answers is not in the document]
fshenton: @pookage what's the big advantage to doing it this way? i.e. using width and height as aspect ratio.

Is the result not the same as if we're using em to define w/h? it all scales no?

src/components/Character/styles.scss
```diff
@@ -1,19 +1,134 @@
...
+			}
+
+			.next {
+				right: var(--padding_container);
+
+				&:after {
+					display: inline-block;
+					content: "";
+					width: 1.36em;
+					height: 2em;
```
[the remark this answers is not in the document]
fshenton: It's one of those things that works so I keep doing it 🙈

src/components/Character/styles.scss
```diff
@@ -1,19 +1,134 @@
...
+		}
+
+		.hero {
+			position: fixed;
+			top: 0;
+			left: 0;
+			width: 100%;
+			height: 100%;
+			z-index: -1;
+			background-color: crimson;
```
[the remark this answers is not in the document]
fshenton: 👨

src/components/Character/styles.scss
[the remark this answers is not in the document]
pookage: Yup, it all scales - but the underlying rules become much more apparent, which is the goal - for example, in this situation it's clear that the ratio is 1:1.5, which `1.36:2` just dosn't make clear.

src/components/Character/Character.jsx
```diff
@@ -55,28 +61,35 @@ export default function Character(props){
...
+						>
+							{ name }
+						</h1>
+						<h2 className={ s.nickName } >
+							{ nickname }
+						</h2>
+						<h3 className={ s.actorName } >
+							{ actor }
+						</h3>
+						<p className={ s.bio } >
```
pookage: when the renderBody updates are merged, don't forget to convert this to a `<div>` and render the `<p>`s as children instead it

src/components/Character/styles.scss
```diff
@@ -1,19 +1,147 @@
...
 		}
-		&.previous {
-			font-size: 1em;
+	
+			.text {
+				overflow-y: scroll;
+			}
+
+			.name {
+				margin-bottom: 0.4em;
```
pookage: all these margins under the elemnts here seem to vary widly - you sure the margins are related to the size of the elements themselves, or are they all actually based on `rem`? Do you get cleaner numbers if you go through the body size?

src/components/Character/Character.jsx
```diff
@@ -55,28 +61,35 @@ export default function Character(props){
...
-				</p>
+			<article className={ s.container }>
+				<div className={ s.content }>
+					<div className={ s.text }> 
+						<h1 className={ s.name } 
+							id={`character__${id}`}
+						>
+							{ name }
+						</h1>
+						<h2 className={ s.nickName } >
```
pookage: can we make this `nickname` to match the variable being used underneath - seems weird to have multiple naming conventions happening here.

src/components/Character/Character.jsx
```diff
@@ -55,28 +61,35 @@ export default function Character(props){
...
+							{ nickname }
+						</h2>
+						<h3 className={ s.actorName } >
```
pookage: can we use `s.actor` here - again, to create some consistancy with the naming we're using for our content variables.

src/components/Character/Character.jsx
pookage: also, probs call that `s.body` innit.

src/components/Character/renderUtils.jsx
```diff
@@ -8,7 +8,7 @@ import {
 import { s } from "./";
 
 
-function renderButton(name, direction){
+function renderButton(name, direction, newIndex){
```
pookage: If we're going to just give it the `index` then there's no point in having the `direction` parameter and we might as well just pass a `label` or either "next" or "previous" - let's either move the index calculations into here and use direction, or remove the direction and add label.

src/components/Character/styles.scss
```diff
@@ -1,19 +1,134 @@
...
+			.button {
+				display: flex;
+				position: absolute;
+				align-items: center;
+				text-decoration: none;
+				--button_margin: 1.2em;
+			}
+
+			.previous {
+				left: var(--padding_container);
```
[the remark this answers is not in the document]
pookage: Move the `.controls` into the `.content` and tehn you won't need any of theeese.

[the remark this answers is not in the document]
pookage: also, might want to:

```
.button {
    &.previous {
      ...
    }
    &.next {
        ...
    }
}
```
to indicate that they're the buttons - as 'previous' and 'next' are generic enough to maybe warrant a bit more specificity.

src/components/Character/styles.scss
```diff
@@ -1,19 +1,147 @@
...
+
+				&::before {
+					content: "";
+					margin-right: var(--button_margin);
+				}
+			}
+
+			.next {
+				right: var(--padding_container);
```
pookage: nahhhhh, use the size of the `content` to do this

src/components/Character/styles.scss
```diff
@@ -1,19 +1,147 @@
...
+					background-color: lime;
+				}
+			}
+
+			.previous {
+				left: var(--padding_container);
+
+				&::before {
+					content: "";
+					margin-right: var(--button_margin);
```
pookage:
```suggestion
```
Whatever this is doing is definitely something it shouldn't - kill these and find another way, haha. I think moving the `.controls` inside your `.content` is going to solve most of your problems here.

src/components/Character/styles.scss
```diff
@@ -1,19 +1,147 @@
...
+				}
+			}
+
+			.next {
+				right: var(--padding_container);
+
+				&::after {
+					content: "";
+					margin-left: var(--button_margin);
```
pookage: kill it with fire.
```suggestion
```

src/components/Character/styles.scss
```diff
@@ -1,19 +1,147 @@
...
+
+// Tablet portrait
+@media (min-width: 768px) {
 
+}
 
+// Tablet landscape
+@media (min-width: 768px) and (orientation: landscape) {
+	.container {
+		width: 50vw;
```
pookage:
```suggestion
		width: 50%;
```

src/components/Character/styles.scss
```diff
@@ -1,19 +1,147 @@
...
+
+// Desktop portrait
+@media (min-width: 1200px) and (max-aspect-ratio: 8/9) {
+
+}
+
+// Desktop landscape
+@media (min-width: 1200px) {
+	.next {
+		right: 15em;
```
pookage:
```suggestion
```

src/shared/fonts.scss
```diff
@@ -9,3 +9,7 @@
 html {
 	font-size: 16px;
 }
+
+h1, h2, h3 {
+	line-height: 1;
```
pookage: This is something for `reset.scss` rather than `fonts.scss`

src/components/Character/Character.jsx
fshenton: will do and done :)

src/components/Character/renderUtils.jsx
fshenton: Dooone

src/components/Character/Character.jsx
```diff
@@ -36,17 +36,23 @@ export default function Character(props){
 	const mask = { clipPath: `url(${maskPath})` };
 
 
-	// next & prev button labels
-	const prevCharacter = characterData[index - 1] || {};
-	const nextCharacter = characterData[index + 1] || {};
+	// wrap indexes if less than 0 or greater than number of characters
+	const prevIndex = index - 1 < 0 ? characterData.length - 1 : index - 1;
+	const nextIndex = index + 1 === characterData.length ? 0 : index + 1;
```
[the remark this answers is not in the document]
fshenton: Went for option A. Nice and readable.

src/components/Character/Character.jsx
```diff
@@ -55,29 +54,29 @@ export default function Character(props){
...
+					>
+						{ name }
+					</h1>
+					<h2 className={ s.nickname } >
+						{ nickname }
+					</h2>
+					<h3 className={ s.actor } >
+						{ actor }
+					</h3>
+					<p className={ s.body } >
```
pookage: Have I already flagged the `<div clasName={s.body}>` and `RENDER.body` for here?

src/components/Character/renderUtils.jsx
```diff
@@ -20,31 +20,28 @@ function renderButton(name, direction){
...
 
+
+	//RENDER
+	// ---------------------------------
+	const safeName       = UTILS.convertToSafeString(name);
+
 	return (
 		<Link
-			className={ `${s.button} ${direction > 0 ? s.next : s.previous}` } 
+			className={ `${s.button} ${label === "Next" ? s.next : s.previous}` }
```
pookage: could we do something here like 
```
const safeLabel = generateSafeString(label);
const className = s[safeLabel];
```

Otherwise we're basically still doing `direction` stuff, but under a guise

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
-		&.next {
-			font-size: 1em;
+
+	.container {
+		position: relative;
+		z-index: 1;
+	}
+
+		.content {
+			position: relative;
```
pookage: is this being used?

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+					overflow-y: scroll;
+				}
+
+			.controls {
+				bottom: var(--button__bottom_spacing); //GUESS
+			}
+
+			.content::after {
+				position: fixed;
+				bottom: var(--button__bottom_spacing); //GUESS
```
pookage: Not sure this is doing what you want it to do? This is creating a big space at the bottom and running the fade all across the width of the device? If we do

```
position: absolute;
bottom: 0;
```
Then it will be constrained to the width of `.content` and it won't break out of the flow like this...

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+}
+
+// Tablet landscape
+@media (min-width: 768px) and (orientation: landscape) {
+	.wrapper {
+		justify-content: center;
+	}
+
+		.container {
+			--button__bottom_spacing: 15vw; //GUESS
```
pookage: yeah, this is an indication of something going wrong - it's a clever use of CSS variables, but it's still essentially a magic number; it would be better to use a solution like the `position: absolute` one mentioned above that doesn't break it out of flow - currently this looks a bit broken:
image
vs
image

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+		}
+}
 
+// Desktop portrait
+@media (min-width: 1200px) and (max-aspect-ratio: 8/9) {
```
pookage: best to remove any queries you're not using 👍  - if you need to remember how to do'em, you can always google later. Let's keep the code clean I reckon.

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+	.name {
+		font-size: 8rem;
+	}
 
+	.nickname {
+		font-size: 3.25rem;
+	}
+
+	.body {
+		max-width: 30vw; // GUESS
```
pookage:
```suggestion
```
I think this is too specific for a child this low - and also setting a max-width as a proportion of the width kinda feels like the answer here is just setting
```
width: 100%
```
and then
```
max-width: 30%
```
on a parent container higher-up somewhere...which I don't think is the solution; I think if there is a max-width going on here then it should probably be in `em` or `rem` to indicate you're targetting a specific number of words-per-line.  If that is the case, though, it feels like it should probably be on `.content`, not here on the `.body`.

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+// Desktop portrait
+@media (min-width: 1200px) and (max-aspect-ratio: 8/9) {
+	
+}
 
+// Desktop landscape
+@media (min-width: 1200px) {
+
+	.name {
+		font-size: 8rem;
```
pookage:
```suggestion
		font-size: 8em;
```
!! 🚨 unnecessary `rem` 🚨 !!
This is definitely `em` as it's this size because it's proportion of the body - if you changed the `font-size` of the `.wrapper` then you'd definitely want this one to scale proportionately instead of staying the same size

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+	}
 
+	.nickname {
+		font-size: 3.25rem;
```
pookage:
```suggestion
		font-size: 3.25em;
```
!! 🚨 unnecessary `rem` 🚨 !!
This is definitely `em` as it's this size because it's proportion of the body - if you changed the `font-size` of the `.wrapper` then you'd definitely want this one to scale proportionately instead of staying the same size

src/components/HeroImage/styles.scss
```diff
@@ -0,0 +1,71 @@
...
+
+}
+
+// Desktop portrait
+@media (min-width: 1200px) and (max-aspect-ratio: 8/9) {
+
+}
+
+// Desktop landscape
+@media (min-width: 1200px) {
```
pookage: let's clean-up these unused media queries and just write'em as we need them from here on out 👌

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
+// Mobile portrait
 .wrapper {
-	font-size: 1em;
+	display: flex;
+	position: absolute;
```
pookage: let's move this `position` down to the `top` and `left` and move that `padding` up to join the `width` and `height`

src/components/Character/styles.scss
```diff
@@ -1,19 +1,189 @@
...
+			    rgba(237, 20, 61, 1) 100%	
+			);
+		}
+
+		.controls {
+			display: flex;
+			position: sticky;
+			width: 100%;
+			max-width: 25rem; // GUESS
+			bottom: 10vw; // GUESS!
```
pookage: maaate, what's going on with these CSS declaration orders? you've been nailing it up to now! 😂

fshenton: Huh? 🤔 We've been doing this no?
Display
Position
Width/Height
Top/Right/Bottom/Left

src/components/Character/Character.jsx
fshenton: Yeah i was struggling to bring over the render util to do it.
Managed it today though, and updated all of this :)

src/components/Character/renderUtils.jsx
fshenton: Yarp good shout, was being a bit lazy I think !

src/components/Character/styles.scss
fshenton: I don't think so!

src/components/Character/styles.scss
fshenton: Yarp, keep thinking of doing it :P

src/components/Character/styles.scss
fshenton: Went with 35em max-width on content :)

src/components/Character/styles.scss
fshenton: 😱

src/components/Character/styles.scss
fshenton: I did what you suggested in the other one, and I feel dirty 🤧

src/components/Character/styles.scss
fshenton: I didn't really get any relationships that I could use :(

src/components/Character/styles.scss
pookage: naww, remember, it's :
- inherited (flex-grow, shrink, basis, grid-columns etc) 
- box model (display, height, width, paddings, margins) 
- position (relative, fixed etc ; top, right, bottom, left, z-index) 
- layout (flex direction, align items, justify, text-align) 
- foreground (color, font weight, text transform, text shadow etc) 
- background (background color, image, size, box-shadow, repeat etc) 
- cursor (pointer events, cursor icon etc) 
transforms (translate rotate etc etc) 
- filters (blur, drop-shadow, brightness, saturation) 
- transitions 
- animations

fshenton: There are many places I haven't been doing that order, and not just in this project loool 🙊

fshenton: I'll keep an eye out going forward!

pookage:
Ahhh, crap, ah well, all goooood. As long as you're consistant!

src/components/HeroImage/HeroImage.jsx
```diff
@@ -0,0 +1,28 @@
...
+			mask: maskPath,
+			description
+		}
+	} = props;
+
+	// inline style to prevent the need for a huge png asset
+	const mask = { clipPath: `url(${maskPath})` };
+
+	return (
+		<div className={ s.hero } id="hero">
```
pookage:
```suggestion
		<div className={ s.wrapper} id="character__hero__ragnar_lothbrok"> 	
```
'hero' is a bit generic; we might have other hero images; let's namespace to `character__hero` and also create a `__character_name` identifer on a per-image basis.

src/components/HeroImage/HeroImage.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+	return (
+		<div className={ s.hero } id="hero"> 	
+			<img 
+				className={ s.image }
+				src={ src } 
+				alt={ description }
+				style={ mask }
+			/>
+			<div className={ s.glow } id="glow"></div>
```
pookage:
```suggestion
			<div className={ s.glow } id="character__hero__glow"></div>
```
Again - `glow` is too generic a namespace, and we already know that we're going to have multiple to need to namespace them.

src/components/HeroImage/styles.scss
```diff
@@ -0,0 +1,71 @@
...
+	.glow {
+		position: absolute;
+		width: 1em;
+		height: 1em;
+		top: 0;
+		left: 50%;
+		font-size: 100vw;
+
+	}
+		.glow::before {
```
pookage:
```suggestion
		&::before {
```
and move it into the `.glow` definition above

src/components/HeroImage/styles.scss
```diff
@@ -0,0 +1,71 @@
...
+			}
+}
+
+// Tablet portrait
+@media (min-width: 768px) {
+
+}
+
+// Tablet landscape
+@media (min-width: 768px) and (orientation: landscape) {
```
pookage: lose the stray media queries I rexxx

src/components/Character/Character.jsx
```diff
@@ -55,29 +58,29 @@ export default function Character(props){
...
+						<div className={ s.body } > 
+							{ body }
+						</div>	
+				</div>
+				<nav className={ s.controls }>
+					{ prevButton }
+					{ nextButton }
+				</nav>	
 			</article>
+			<HeroImage image={ image } />
```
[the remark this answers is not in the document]
fshenton: As i couldn't just pass src by itself (need mask and description also), i've refactored so we now pass three props to HeroImage. Think it's OK.
Bit unsure about the default destructure in Character though. Do we need to do some sort of selective rendering if the image doesn't exist / is an empty object?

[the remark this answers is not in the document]
fshenton: @pookage

src/components/Character/styles.scss
```diff
@@ -1,19 +1,184 @@
...
-			font-size: 1em;
+
+	.container {
+		position: relative;
+		z-index: 1;
+	}
+
+		.content {
+			margin-top: 40vh;
+			max-width: 35em;
```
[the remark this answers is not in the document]
fshenton: Ahh, i couldn't work this out. Thanks!

src/components/Character/styles.scss
```diff
@@ -1,19 +1,184 @@
...
+
+			.button {
+				display: flex;
+				align-items: center;
+				text-decoration: none;
+				--button_margin: 1em;
+
+				&::before, 
+				&::after {
+					display: inline-block;
```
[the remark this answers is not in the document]
fshenton: Can't do this as it'll show before + after for BOTH prev and next.
i.e. there'll be 4 lovely lime green boxes, instead of two.
I'll do the best i can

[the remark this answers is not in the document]
fshenton: &.prev::before,
&.next::after {
	content: "";
}

😏

src/components/Character/styles.scss
```diff
@@ -1,19 +1,184 @@
...
+}
+
+// Tablet landscape
+@media (min-width: 768px) and (orientation: landscape) {
+	.wrapper {
+		justify-content: center;
+	}
+
+		.container {
+			--button__bottom_spacing: 15vw; //GUESS
```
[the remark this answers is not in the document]
fshenton: so aggressive and violent 😨

src/components/HeroImage/HeroImage.jsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { s } from "./";
+
+export default function HeroImage(props){
+
+	const {
+		image: {
```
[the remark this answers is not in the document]
fshenton: Cool. I was trying to do it the other way around! 😛

src/components/Character/styles.scss
```diff
@@ -1,19 +1,184 @@
...
+				&.next {
+					&::after {
+						content: "";
+						margin-left: var(--button_margin);
+					}
+				}
+			}
+
+// Mobile landscape
+@media (orientation: landscape) {
```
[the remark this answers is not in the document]
fshenton: Ok I managed to do it, but I used translate which is probably a bit sketch. Don't hurt me pls 🙈

src/components/Character/Character.jsx
```diff
@@ -4,7 +4,10 @@ import {
 	CharactersContext as Characters,
 	ACTIONS
 } from "COMPONENTS/Characters/";
+import HeroImage from "COMPONENTS/HeroImage/";
 import { RENDER, s } from "./";
+import RENDER_TEXT from "SHARED/renderUtils.jsx";
```
pookage: Not sure about this naming - `RENDER` works as namespace but `RENDER_TEXT` implies its a value rather than an interface - if you want to namespace like this it might be worth doing something like `RENDER.TEXT` instead, although that probably implies that your structure is too complex. I think we might as well just stick to `RENDER` here instead.

src/components/Character/Character.jsx
```diff
@@ -19,34 +22,41 @@ export default function Character(props){
...
 	const { name: prevCharName } = prevCharacter;
 	const { name: nextCharName } = nextCharacter;
 
+	const prevButton = RENDER.button(prevCharName, "prev", prevIndex);
+	const nextButton = RENDER.button(nextCharName, "next", nextIndex);
 
 	// active character checks
 	const hidden = index !== activeIndex;
 
+	const body = RENDER_TEXT.body(bodyData, {
```
pookage: as above - `RENDER.body` is better, I think, as it shows you're using the `RENDER` interface - this just looks like a weirdly named function or class or something?

src/components/Character/Character.jsx
```diff
@@ -19,34 +22,41 @@ export default function Character(props){
...
+		scope: "character", //for key names
+		className: s.paragraph
+	});
+
+	const idString = UTILS.convertToSafeString(`${name} ${nickname}`);
```
pookage: to give an example comparisson with what I'm talking about - wouldn't it be weird if this was `UTILS__STRING.convertToSafe`? You get me?

src/components/Character/Character.jsx
```diff
@@ -55,29 +65,35 @@ export default function Character(props){
...
+						{ body }
+					</div>	
+				</div>
+				<nav className={ s.controls }>
+					{ prevButton }
+					{ nextButton }
+				</nav>	
 			</article>
+			<HeroImage 
+				idString={ idString }
```
pookage:
```suggestion
				id={ heroId }
```
I don't think an `id` can be anything tother than a string, so maybe a superfluous name here, and `idString` as a value doesn't say much about what it is - so `heroId` works better imho.

src/components/Character/styles.scss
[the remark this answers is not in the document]
pookage: ahhh, good shout, and good solution.

src/components/Character/styles.scss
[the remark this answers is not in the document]
pookage: Fraid so! Big no-no there. It kinda works, but remember that the trick with CSS isn't in getting it to work, it's to have the CSS say what you mean.  We don't mean "wherever the body text is - move it to the right", we mean "lay out the title and the body side-by-side".  It's not actually too painful.  Here's what we have right now:

```
<div class="character__content">
    <h1>Ivar</h1>
    <h2>The Boneless</h2>
    <h3>Alex Hogh Andersen</h3>
    <div class="character__body">
        <p />
        <p />
    </div>
</div>
```

Now, what we want is for the 3 headings to be side-by-side with the body, right? SO, all we need to do is wrap the `<h*>` elements inside a `<div>` and then make `character__content` `display: flex;` and `flex-direction: row` - then you're done!   If you didn't want to add any new `<div>` elements, then you could also use `display: grid` set-up a 2-col 3-row grid and make the body span all the rows.

It was good that you identified that `translate` here was the wrong shout - but the next step there is to take a step back and re-examine the fundamentals of the problem rather than whack-in a solution.  This is a layout problem, and translate is for offsets to occur after the layout, so there's your red flag.

Give's a holla if you need a hand with this!

src/components/Character/styles.scss
```diff
@@ -129,25 +129,21 @@
...
 			.content {
-				margin-top: 0;
+				margin-top: 60px;
 			}
 
 				.body {
 					height: 25vh; //GUESS
 					margin-bottom: 2em; //GUESS
-					overflow-y: scroll;
+					transform: translate(100%, -200%);
```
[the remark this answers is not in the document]
pookage:
```suggestion
```
See above comment for instructions on how to do this, but yeah - that translate deffo needs to gooooo.

src/components/HeroImage/HeroImage.jsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { s } from "./";
+
+export default function HeroImage(props){
+
+	const {
+		idString,
```
pookage:
```suggestion
		id,
```
As mentioned in other comment - `idString` is a bit superflous - kinda like saying `listArray` or `countNo`

src/components/Character/styles.scss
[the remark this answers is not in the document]
fshenton: Cool yeah, i was trying to do it without using a new div. I have an irrational fear of using them! xD

src/components/Character/Character.jsx
fshenton: Did this, as discussed:
 const RENDER = { ...RENDER_A, ...RENDER_B };

Works fine. Using A and B as i don't know of a naming convention that sidesteps the verb -> noun issue !

pookage: Will do a CR tonight / tomorrow, but as for the verb -> noun issue, instead of `RENDER_TEXT`, you can just do `TEXT_RENDER` ;)

src/components/Character/Character.jsx
```diff
@@ -15,38 +18,47 @@ export default function Character(props){
 
 	//RENDER
 	//----------------------------
+	const RENDER = { ...RENDER_A, ...RENDER_B };
```
pookage:
```suggestion
	const RENDER = { ...CHARACTER_RENDER, ...SHARED_RENDER };
```
Just to make it somewhat more useful in terms of naming

src/components/Character/Character.jsx
```diff
@@ -55,29 +67,37 @@ export default function Character(props){
...
+						{ body }
+					</div>	
+				</div>
+				<nav className={ s.controls }>
+					{ prevButton }
+					{ nextButton }
+				</nav>	
 			</article>
+			<HeroImage 
+				heroId={ heroId }
```
pookage:
```suggestion
				id={ heroId }
```
Feel like I might have mentioned this in another one - apologies for missing it if not - we already have 'hero' in the component name so don't really need to say `heroId` as the name of the prop

src/components/Character/styles.scss
```diff
@@ -1,19 +1,231 @@
...
+					background-color: lime;
+				}
+
+				&.prev::before,
+				&.next::after {
+					content: "";
+				}
+
+				&.prev {
+					&::before {
```
pookage: If nothing else is happening in `&.prev` then we can just do `&.prev::before`

src/components/Character/styles.scss
```diff
@@ -1,19 +1,231 @@
...
+						margin-right: var(--button_margin);
+					}
+				}
+
+				&.next {
+					&::after {
```
pookage: as above - we can just merge this to `&.next::after`

src/components/Character/Character.jsx
fshenton: You did mention it before but I thought using id in HeroImage would cause issues. But it appears that id is just treated like any other prop?

PR #45, opened 2019-10-03
Feature/layout/episode

fshenton:
Just getting started really. But in case you wanted to have an early look :)


src/components/Episode/Episode.jsx
```diff
@@ -53,6 +44,17 @@ export default function Episode(props){
 					thumbnail={ thumbnail }
 					trailer={ trailer }
 				/>
+				<div className={ s.content }>
+					<h2 className={ s.episode }>
```
pookage: while technically having the `<h2>` before the `<h1>` is fine, maybe we should just make `.content` `display: flex` and have

```
.episode {
    order: 1;
}
.title {
   order: 2;
}
.body {
   order: 3;
}

File: src/components/Episode/Episode.jsx [hunk 2]
```diff
@@ -53,6 +44,17 @@ export default function Episode(props){
...
+						Episode { number }
+					</h2>
+					<h1 className={ s.title }>
+						{ title }
+					</h1>
+					<p className={ s.body }>
```
pookage:
```suggestion
					<div className={ s.body }>
```
As we've done on the others - let's make the body container its own thing so that we can call `RENDER.body` and fill it with multiple `<p>` tags if needs-be.

src/components/Episode/styles.scss
```diff
@@ -1,7 +1,31 @@
 .wrapper {
-	font-size: 1em; //placeholder styling
-
+	margin-top: 20vh;
```
pookage: Is this a margin on the top of this component, or a padding on the top of its parent? Haven't checked, I'm just asking in-case you haven't considered it yourself yet!

src/components/Episode/styles.scss
```diff
@@ -1,7 +1,31 @@
...
+	position: absolute;
+	
 	&[aria-hidden="true"]{
 		opacity: 0;
 	}
-}
\ No newline at end of file
+}
+
+	.content {
```
pookage: is this a `.container`?

src/components/Episode/styles.scss
```diff
@@ -1,7 +1,31 @@
...
+		padding: 10%;
```
pookage: is this actually `%` or is it `em`?

src/components/Episode/styles.scss
```diff
@@ -1,7 +1,31 @@
...
+		}
+
+		.title {
+			font-size: 1.7em;
+			font-family: var(--font_family__heading);
+		}
+
+		.body {
+			margin-top: 1.5em;
+			line-height: 1.5em;
```
pookage:
```suggestion
			line-height: 1.5;
```
This is correct, but Edge still fucking shits bricks if you provide a `line-height` that has units - luckily, `1.5` is `1.5em` so we're good to continue. Just an annoying quirk to memorise I'm afraid - although this may change now that Edge is a Chromium build.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -91,18 +91,26 @@ export default function EpisodePreview(props){
...
-				aria-controls={ previewId }
-				aria-label={ `Watch preview for episode ${episodeNo}: ${title}.` }
-				onClick={ openOverlay }
-			>
-				<img
-					src={ thumbSrc }
-					alt={ thumbDesc }
-				/>
-			</a>
+		<aside className={ s.container }>
```
pookage:
```suggestion
		<aside className={ s.wrapper}>
```
Outermost element _and_ elastic width; deffo a wrapper 😘

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -91,18 +91,26 @@ export default function EpisodePreview(props){
...
+					<img
+						src={ thumbSrc }
+						alt={ thumbDesc }
+					/>
+					<svg className={ s.playButton } 
+						 viewBox="0 0 200 200" 
+						 alt="Play video">
+	                    <circle cx="100" cy="100" r="90" fill="none" strokeWidth="15" stroke="#fff"/>\
+	                    <polygon points="70, 55 70, 145 145, 100" fill="#fff"/>\
+                	</svg>
```
pookage: Good use of SVG, but they tread this weird line of content / asset - so what I tend to do with these is put them into `SHARED/icons/` and do the usual interface stuff w/index.js :

```
import Play from './Play.jsx';
import Stop from './Stop.jsx';

export { Play, Stop };
```

As long as you make it so that you can use them by doing something like:

```
import { Play as PlayIcon } from "SHARED/icons/";

return (
    <PlayIcon />
);
```
and maybe add support to change their colour as a prop?
```
<PlayIcon color="#ff0000" />
```

src/components/EpisodePreview/styles.scss
```diff
@@ -1,5 +1,39 @@
...
+		}
+	
+			img { //need classname
+				width: 100%;
+				height: 100%;
+			}
+
+			svg { //need classname
+				width: 17.5%;
+				height: 25%;;
```
pookage: I think SVGs have scaling built into them, so you can specify the general size you'd like them to be contained within and they'll scale themselves up as much as possible whilst maintaining their aspect ratio...although you might need to put `preserveAspectRatio="true"` on them, or something? Worth looking-into and having a fiddle.

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -40,21 +40,23 @@ export default function EpisodesNav(){
...
-				onClick={ prevEpisode }
-			/>
-			<button 
-				className={ s.navButton }
-				aria-label="Next episode." 
-				aria-controls="episodes__items episodes__pagination"
-				aria-hidden={ isLastEpisodeActive }
-				onClick={ nextEpisode }
-			/>
+		<nav className={ s.controls }>
```
pookage:
```suggestion
		<nav className={ s.wrapper}>
```

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -40,21 +40,23 @@ export default function EpisodesNav(){
...
+			<div className={ s.wrapper }>
```
pookage:
```suggestion
			<div className={ s.container}>
```

src/components/EpisodesNav/EpisodesNav.jsx
pookage: `NavControls__controls` seems excessive 😛

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -40,21 +40,23 @@ export default function EpisodesNav(){
...
+				<button 
+					className={ s.navButton }
```
pookage:
```suggestion
					className={ s.button}
```
No need to namespace - you're already in a component that has 'nav' in the name.

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -40,21 +40,23 @@ export default function EpisodesNav(){
...
+					aria-label="Previous episode."
+					aria-controls="episodes__items episodes__pagination"
+					aria-hidden={ isFirstEpisodeActive }
+					onClick={ prevEpisode }
+				/>
+				<button 
+					className={ s.navButton }
```
pookage:
```suggestion
					className={ s.button }
```

src/components/Episode/Episode.jsx
fshenton: Yah, that's nice. Forgot flex could do that

src/components/Episode/Episode.jsx
fshenton: done!

fshenton: Pulled over the most up to date renderBody function from layout, and also corrected a small issue where an empty anchor would be included if the findLink function returned false.

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Had to change another div className to content as there would have been a conflict with two wrappers.  Not sure if content is right here, hmmmm

src/components/Episode/styles.scss
fshenton: Did some more calculations and actually the best value (that applies pretty closely over mobile and desktop) is 10vw.

src/components/Episode/styles.scss
fshenton: Yeah that does make more sense for it to be padding-top for parent 👍

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: Done!

src/components/EpisodePreview/styles.scss
fshenton: Made some changes, not perfect but better!

fshenton:
Notes:
- vertical centering of pagination + prev/next buttons is going out of whack for some reason. Can't consistently repro the issue though. Very weird
- ran into issues again with scrolling for tablet portrait
- using next/prev causes width of preview/text to change 🤷‍♂ 
- hacked the text container position in desktop fullscreen (getting tired) and struggled with pagination positioning

pookage:
Before looking at the code I did a quick testing the pass on this one and have a couple general comments:

### Mobile Landscape
- `EpisodePreview` should be fixed / sticky, I think; it feels weird for it to scroll when it's not revealing anything underneath itself
- I think the pagination would feel more natural in vertical on the right here, instead of along the bottom; it buys back some limited real-estate
- We can scroll onto a white background; is that a problem or will that be okay when assets are in?

### Portrait Desktop

This feels a bit weird - we might need to come up with a specific layout plan for this one; or it might be a matter of just triggering the full-desktop layout a bit sooner.

image

src/components/EpisodesNav/EpisodesNav.jsx
```diff
@@ -40,24 +45,39 @@ export default function EpisodesNav(){
...
+					aria-hidden={ isLastEpisodeActive }
+					onClick={ nextEpisode }
+				/>
+				{ isSmall && 
+					<ol 
+						className={ s.pagination }
+						id="episodes__pagination"
+					>
+						{ pagination }
+					</ol>
```
pookage: It's the same element here and in the other place, so rather than having to update in two places, we should move this into the `renderUtils` I reckon `RENDER.pagination` should refer to this `<ol>`  instead - it can call `renderPaginationLink` to generate its children inside the utils.

pookage: to clarify, that means we can do
```
const pagination = RENDER.pagination(data);

...

{isSmall && pagination}
...
{!isSmall && pagination}
```

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -37,10 +37,28 @@ function renderPaginationLink(data, index){
...
-	const isActive = index === activeIndex;
+	const isActive 	= index === activeIndex;
+
+	let opacity = 1;
+	if(!isActive){
+		const decayDistance = 4;
+		const decayRate     = 1/decayDistance;
+		const fadeOffset    = Math.abs(index - activeIndex);
+		
+		opacity -= fadeOffset * decayRate;
```
pookage:
```suggestion
		opacity = 1 - (fadeOffset * decayRate);
```
as sexy as `-=` is, it implies that this function is going to be iterated over in a local loop or something, so I think keeping it pure assignment might be better.

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -37,10 +37,28 @@ function renderPaginationLink(data, index){
 
 	const episodeId = UTILS.convertToSafeString(title, "-");
 
-	const isActive = index === activeIndex;
+	const isActive 	= index === activeIndex;
+
+	let opacity = 1;
+	if(!isActive){
```
pookage: For the purposes of A/B-ing it, I think it works better to be like

```
let opacity;
if(isActive) opacity = 1;
else {
    ...
    opacity = 1 - (fadeOffset  * decayRate);
}
```

rather than have a default value and the secretly overriding it in an if statement. by putting a `let` above an `if/else` it makes it super-clear what the role of that `if/else` is, and also that the `opacity` is about to be definied right here and now.

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -37,10 +37,28 @@ function renderPaginationLink(data, index){
...
+		const fadeOffset    = Math.abs(index - activeIndex);
+		
+		opacity -= fadeOffset * decayRate;
+	}
+	
+	const offset = index - activeIndex;
+
+	const style = {
+		transform: `translateX(${offset}em)`,
+		opacity: opacity
```
pookage:
```suggestion
		opacity
```
If the key / value names are the same then you don't need to specify both sides

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -37,10 +37,28 @@ function renderPaginationLink(data, index){
...
+	};
 
 	return (
 		<li 
+			className={ isActive ? s.active : s.inactive }
```
pookage:
```suggestion
			className={`${s.item} ${isActive ? s.active : s.inactive}`}
```
I guarantee I'm going to see some funkiness below if there's not a shared className for both states, haha.  The reason we do this is that we want to create a 'rest' / 'shared' state and then define what is different FROM that state with the modifying states, if that makes sense?

src/components/EpisodesNav/styles.scss
```diff
@@ -1,14 +1,120 @@
...
+	bottom: 5%;
+	transform: translateX(-50%);
 }
 
-.navButton {
-	width: 50px;	//placeholder so they are visible
-	height: 50px;
-	margin: 5px;
+	.container {
+		display: flex;
```
pookage:
```suggestion
		display: flex;
		position: relative;
		z-index: 1;
```
I found some z-fighting when I was testing and couldn't click the buttons.

src/components/EpisodesNav/styles.scss
```diff
@@ -1,14 +1,120 @@
...
+		}
+
+			.inactive, .active {
+				display: flex;
+				position: absolute;
+				height: 100%;
+				left: 50%;
+				align-items: center; 
+
+				a {
```
pookage:
```suggestion
				.paginationLink  {
```

src/components/EpisodesNav/styles.scss
```diff
@@ -1,14 +1,120 @@
...
+		.pagination {
+			order: 2;
+			display: flex;
+			width: 40%;
+			height: 1.5em;
+			align-items: center;
+			list-style: none;
+		}
+
+			.inactive, .active {
```
pookage:
```suggestion
			.item {
```

aaah, here's the fellah! yeah, let's just use the 'rest' state for this one with a separate class name - that way we can do `item.active` and `item.inactive` which reads better imho

src/components/EpisodesNav/styles.scss
```diff
@@ -1,14 +1,120 @@
...
+					font-size: 1.15em;
+					font-weight: normal;
+					pointer-events: none; //prevents overlap with prev/next
+				}
+
+				.active {
+					font-size: 1.5em;
+					font-weight: bold;
+					pointer-events: all;
+				}
```
pookage: We can do `&.inactive` and `&.active` and put them as a child of `.item` now that we have that rest state in!

src/components/EpisodesNav/styles.scss
```diff
@@ -1,14 +1,120 @@
...
+		width: 33%;
+	}
+}
+
+//Desktop fullscreen
+@media (min-width: 1200px) {
+	.wrapper {
+		position: fixed;
+		width: 2em;
+		height: 10em;
```
pookage: oop, looks like extra indentation here

src/shared/icons/Play.jsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { s } from "./";
+
+export default function Play(props) {
+
+	const { colour } = props;
+
+	return (
+		<svg className={ s.icon } 
+			 viewBox="0 0 128 128"
```
pookage: See my notes in the main thread here

src/shared/icons/Play.jsx
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { s } from "./";
+
+export default function Play(props) {
+
+	const { colour } = props;
```
pookage:
```suggestion
	const { 
		colour, 
		className = "" 
	} = props;
```
Passing down a className allows us to scale the icon and add animations and stuff if we need to!

src/shared/icons/Play.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+	return (
+		<svg className={ s.icon }
```
pookage:
```suggestion
		<svg className={ `${s.icon} ${className}` } 
```

src/shared/icons/Play.jsx
```diff
@@ -0,0 +1,29 @@
...
+	        	cy="64" 
+	        	r="56.5" 
+	        	fill="none" 
+	        	strokeWidth="15" 
+	        	stroke={ colour }
+	        />
+	        <polygon 
+	        	points="48,48 48,80 80,64" 
+	        	fill={ colour }
+	        />
```
pookage: needs indenting

src/shared/icons/styles.scss
```diff
@@ -0,0 +1,8 @@
+.icon { //need classname
+	width: 128px;
+	height: 128px;
```
pookage:
```
font-size: 128px;
width: 1em;
height: 1em;
```
You know the drill - preserve the aspect ratio, specify size with `font-size`

src/shared/icons/styles.scss
```diff
@@ -0,0 +1,8 @@
...
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	transform: translate(-50%, -50%);
```
pookage: This is specific to the `<EpisodePreview />` component - so shouldn't be defined here, and should be passed down via the `className` prop that I introduced up there ^^

src/shared/renderUtils.jsx
```diff
@@ -1,19 +1,58 @@
...
+	const customisedRenderParagraph = renderParagraph.bind(this, options);
+
+	return paragraphs.map(customisedRenderParagraph);
 }// renderBody
 
-function renderParagraph(text, index){
-	const key = `about__body__paragraph_${index}`;
+
+function renderParagraph(options, text, index){
+	const { scope, className } = options;
```
pookage:
```suggestion
	const { 
		scope, 
		className = ""
	} = options;
```
Good to default `className` to `""` otherwise we end up with elements with loads of `<div class="undefined">` and stuff like that.

src/shared/utils.js
```diff
@@ -4,6 +4,29 @@ function convertToSafeString(string, seperator="_"){
...
+
+function getDeviceHeight(){
+	const h = Math.max(
+		document.documentElement.clientHeight,
+		window.innerHeight
+		|| 0
+	);
+
+	return h;
+}// getDeviceHeight
```
pookage: as mentioned above - could be worth combining these into a `getDeviceDimensions` function. Also boooooo single letter variablesssss

src/components/App/context.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { createContext, useReducer } from "react";
+import { reducer } from "./";
+import UTILS from "SHARED/utils.js";
+
+const w = UTILS.getDeviceWidth(); //just using width to determine viewport size
+
+const initialState = {
+	isSmall:  w < 768 ? true : false,
+	isMedium: w >= 768 && w < 1200 ? true : false,
+	isLarge: w >= 1200 ? true : false,
```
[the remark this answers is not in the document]
fshenton: LOL it appears i've forgotten how to use booleans normally, my mind is plagued with ternaries !

src/components/App/context.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { createContext, useReducer } from "react";
+import { reducer } from "./";
+import UTILS from "SHARED/utils.js";
+
+const w = UTILS.getDeviceWidth(); //just using width to determine viewport size
```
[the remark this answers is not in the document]
fshenton: Cool, first time using useEffect :)

src/components/App/reducer.js
```diff
@@ -0,0 +1,25 @@
...
+
+	switch(type){
+		case ACTIONS.UPDATE_VIEW_SIZE: //Should be called if window is resized (not implemented)
+			const w = UTILS.getDeviceWidth();
+			const h = UTILS.getDeviceHeight();
+			return {
+				...state,
+				isSmall:  w < 768 ? true : false,
+				isMedium: w >= 768 && w <= 1200 ? true : false,
+				isLarge: w >= 1200 ? true : false,
```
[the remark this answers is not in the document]
fshenton: Removed one unnecessary =. 
The others are needed I think. Changing the check values to 767 or 1199 to allow for > or <, means 768 and 1200 (exactly) aren't captured. So we'd need to use <= and >= regardless. Unless i'm missing something!

src/components/EpisodePreview/styles.scss
```diff
@@ -1,5 +1,89 @@
...
+				background-image: url("/path/to/lord-of-the-rings-5.jpg");
+			}
+		}
+	}
+
+	.content {
+		position: absolute; //placeholder
+
+		&[aria-hidden="true"]{
+			opacity: 0;
```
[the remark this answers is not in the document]
fshenton: good idea. Unfortunately I can't search for other uses, so let me know if you spot any issues!

src/components/EpisodesNav/renderUtils.jsx
fshenton: Ah, forgot that

src/components/Episode/styles.scss
```diff
@@ -1,7 +1,128 @@
...
+			.episode {
+				order: 1;
+				margin-bottom: 1.67em;
+				font-size: 1em;
+				font-family: var(--font_family__heading);
+			}
+
+			.title {
+				order: 2;
+				font-size: 1.7em;
```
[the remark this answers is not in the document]
fshenton: you're SO thirsty for that golden ratio!

src/components/EpisodePreview/styles.scss
```diff
@@ -1,5 +1,89 @@
...
+	width: 100%;
+	--thumbnail_width: 100%;
+	--thumbnail_height: 38vh;
+}
+
+	.thumbnail {
+		display: flex;
+		position: relative;
+		width: var(--thumbnail_width);
+		height: var(--thumbnail_height);
```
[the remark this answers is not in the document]
fshenton: Why would it skew? 🤔 probs something i'm not getting!

src/components/EpisodesNav/renderUtils.jsx
fshenton: good shout!

src/components/EpisodesNav/EpisodesNav.jsx
fshenton: Good idea. It didn't sit well with me having two identical lines of code!

src/components/EpisodesNav/styles.scss
fshenton: Hmm, looks ok on mine. lemme know if still an issue!

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -91,21 +92,39 @@ export default function EpisodePreview(props){
...
+					className={ `${s.ghost} ${s.bottom} ${s.left}`}>
+				</div>
+				<div 
+					className={ `${s.ghost} ${s.bottom}` }>
+				</div>
+			</div>
 			<div
 				id={ previewId }
-				className= { s.wrapper }
+				className= { s.content }
```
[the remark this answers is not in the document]
fshenton: yeah it probably should be. Think it's best to do it on the preview branch, to avoid conflicts

src/shared/icons/Play.jsx
fshenton: Looking fine to me. Re-indented anyway, hopefully that fixes.

src/shared/icons/styles.scss
fshenton: cool, makes sense! :)

fshenton:
- left SVG looking janky as i'll need to spend a fair bit of time on that during details pass anyway.
- most of it looks decent. I'm sure there's more polish to be done though

pookage:
Cooool, okay, before I dive into the code, here's my testing pass! Some of it might just be me being derpy, but:

### All Devices

1. We should probably add gesture controls for touch devices - if it's not on the board we should add it for the details / polish pass
2. Pagination orientation doesn't change when you resize between devices until you refresh the browser - the client provider probably needs to add a resize listener and do its thing whenever the viewport size changes
3. The pagination looks kinda small? Probs best to leave that til the details pass tho
4. Permalinking to a specific episode should open on that episode - this is probably going to be some logic in the `/Episodes/context.jsx` to determine the correct `activeIndex` based on the id.  You can get the id from the url using `window.location.hash`
5. Next / Previous should update hash by pushing to the browser history state - we're going to do this more later when we do the routing between pages, but if you want to take a stab at it, it's using the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API).  Don't sweat if not, though - just whack it on the board so we don't forget.
6. Pagination looks like it only has a small height and none of the inactive items are clickable? Probs want to address that, 'cos we wanna be able to click the non-active numbers
7. Probs don't wanna be able to click the active number as we're already there.
8. SVG stuff - you already know and imma probs comment in-code, but it looks like you've specified some points outside of the viewbox, so its croppin'.

### Mobile Landscape

1. Can still scroll to white? Again, just ignore if this isn't actually a problem and that's just asset stuff

### Tablet Portrait

1. Next / Prev are on-top o the pagination instead of either side? This looks like it happens from tablet upwards because the buttons are no longer direct siblings of the pagination?

### Snapped-Half Desktop

1. Video / Text are stacked vertically instead of side-by-side? Doesn't look intentional...

image


### Desktop Landscape

1. Ghosts are triggering vertical scrollbars - we should whack `overflow: hidden` on the `Episode__wrapper` for desktop to prevent this.
2. White bar on side? Again - ignore if just asset stuff.

image

src/shared/icons/styles.scss
```diff
@@ -0,0 +1,12 @@
...
+	width: 1em;
+	height: 1em;
+}
+
+	.preview {
+		position: absolute;
+		top: 50%;
+		left: 50%;
+		font-size: 128px;
+		transform: translate(-50%, -50%);
```
pookage: ahhh, I get what you're doing here, but this definitely belongs in the styles for episode preview - this icon could theoretically be pulled out and used anywhere - in which case it doesn't need to know about preview styles, y'know? Just make this `s.icon` in `/EpisodePreview/styles.scss` and pass that down I reckon.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -91,21 +92,34 @@ export default function EpisodePreview(props){
...
+					onClick={ openOverlay }
+				>
+					<img
+						className={ s.image }
+						src={ thumbSrc }
+						alt={ thumbDesc }
+					/>
+					<PlayIcon 
+						colour="#FFF" 
+						className="preview"
```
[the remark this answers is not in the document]
fshenton: Ah, i wanted to do that but because EpisodePreview didn't have the style, it was undefined. 
Feels a bit weird. But this works 👍

src/components/EpisodesNav/styles.scss
```diff
@@ -1,14 +1,142 @@
...
+}
+
+//Desktop half screen
+@media (min-width: 1200px) {
+	.wrapper {
+		position: fixed;
+		width: 2em;
+		height: 8em;
+		left: 85%;
+		bottom: 25%;
```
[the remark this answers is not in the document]
fshenton: I wish!  😛 

image

src/index.js
```diff
@@ -1,14 +1,14 @@
...
+import Client from "COMPONENTS/Client/";
 import "SHARED/global.scss";
 import "SHARED/reset.scss";
 
 window.addEventListener("DOMContentLoaded", init);
 
 function init(){
 	ReactDOM.render(
-		<App />,
+		<Client />, //Client wraps App
```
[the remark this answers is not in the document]
fshenton: Sweet, this is what I wanted.  Felt wrong using Client instead of App.

src/components/Client/Client.jsx
```diff
@@ -0,0 +1,12 @@
...
+import { ClientProvider } from "./";
+
+export default function Client(){
+
+	return (
+		<ClientProvider>
+			<App />
+		</ClientProvider>
+	);
+}//Client
```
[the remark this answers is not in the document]
fshenton: Yarp, removed this. Thanks!

src/shared/icons/Play.jsx
```diff
@@ -0,0 +1,32 @@
...
+			<circle 
+				cx="64" 
+				cy="64" 
+				r="56.5" 
+				fill="none" 
+				strokeWidth="15" 
+				stroke={ colour }
+			/>
+			<polygon 
+				points="48,48 48,80 80,64"
```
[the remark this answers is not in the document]
fshenton: Ha, yeah i totally just copied it from the net and didn't really know what i was doing with it 😄

src/shared/icons/styles.scss
fshenton: Ok fair enough. I've copied the styles into s.icon for EpisodePreview. Had to specify a height for the div wrapper as it otherwise was taller than it was wide, for no good reason 🤷‍♂

src/components/Client/context.jsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const ClientContext = createContext();
+
+function ClientProvider(props){
+
+	const { children } = props;
+
+	const [state, dispatch] = useReducer(reducer, initialState);
+
```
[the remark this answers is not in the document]
fshenton: Had to move it into App as couldn't access Context->dispatch within context.jsx. 
Makes sense to be in there as I'm already using an effect (on mount) to set the screen size.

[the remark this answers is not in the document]
fshenton: Debouncing looks interesting, though I fear getting stuck on it for a month if I dive in haha

fshenton:
Some bits i've added to the board:
- gesture controls
- episode permalinking from URLs (i don't even)

Not sure how to repro what you're seeing on half-screen desktop. That media break point is more like a media _pain_ point....heh heh heh

The white background stuff is fine as far as i can tell. If there's still something weird once I've plugged in assets, i'll investigate further.

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -37,10 +56,39 @@ function renderPaginationLink(data, index){
...
+		axis   = "Y";
+		factor = 1.5;
+	}
+	else{ 
+		axis   = "X";
+		factor = 1;
+	}
+
+	const style = {
+		transform: `translate${axis}(${factor*offset}em)`,
```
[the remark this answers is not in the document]
pookage: maybe even 
```const translation = `${factor * offset}em`;``` 
so that the transform can just be `${translate}(${translation})`? I dunno. whatever you dig.

File: src/components/PreviewOverlay/PreviewOverlay.jsx [hunk 47]
```diff
@@ -0,0 +1,112 @@
...
+export default function PreviewOverlay(props){
+	
+	// PROPS
+	// ---------------------------
+	const {
+		episodeId,
+		index,
+		episodeNo,
+		title,
+		previewId,
```
pookage: I feel like this one should be `id` - we can always alias it as `previewId` if we want to A/B it, but it feels odd that it doesn't have, like, a 'main' id?

src/components/PreviewOverlay/PreviewOverlay.jsx
```diff
@@ -0,0 +1,112 @@
...
+		trailerSrc,
+		prevEpisode, //bound callback
+		nextEpisode, //bound callback
+		closeOverlay //bound callback
```
pookage: mmmm, I don't think these should be coming through as props? We can use `useContext` in the same way that we did in the other component - that way we can keep this component more modular and stuff. This is also quite a lot of props - are there any more than we can derive from the context instead of passing as a prop?

src/components/PreviewOverlay/PreviewOverlay.jsx
```diff
@@ -0,0 +1,112 @@
...
+	const isLast            = index === episodeCount - 1;
+
+	const { src: prevTrailerSrc } = prevTrailer;
+	const { src: nextTrailerSrc } = nextTrailer;
+
+
+	return (
+		<div
+			id={ previewId }
+			className= { s.container }
```
pookage:
```suggestion
			className= { s.wrapper }
```
🙃

src/components/PreviewOverlay/PreviewOverlay.jsx
```diff
@@ -0,0 +1,112 @@
...
+			aria-hidden={ hidden }
+		>
+			<header className={ s.details }>
```
pookage:
```suggestion
			<header className={ s.container}>
```
🐵

src/components/PreviewOverlay/PreviewOverlay.jsx
```diff
@@ -0,0 +1,112 @@
...
+							<div className={ `${s.bar} ${s.bar1}` } />
+							<div className={ `${s.bar} ${s.bar2}` } />
+						</div>
+					</a>
+				</nav>
+			</header>
+			<video className={ s.video }>
+				{ isOverlayActive && (
+					<source 
+						src={trailerSrc}
```
pookage: now that we have resolution controls via the Client provider, we can start to think about serving assets at the appropriate resolutions - ie. stuff like:

```
const data = {
    preview: {
        poster: "path/to/poster.jpg",
        sources: {
            webm: {
                low: "path/to/preview/01/webm/low.webm,
                medium: "path/to/preview/01/webm/medium.webm,
                high: "path/to/preview/01/webm/high.webm
            },
            mp4: {
                low: "path/to/preview/01/mp4/low.mp4,
                medium: "path/to/preview/01/mp4/medium.mp4,
                high: "path/to/preview/01/mp4/high.mp4
            },
        }
    }
};

let resolution;
if(isSmall)       resolution = "low";
else if(isMedium) resolution = "medium";
else              resolution = "high";

const webm = data.preview.sources.webm[resolution];
const mp4  = data.preview.sources.mp4[resolution];

return (
    <video poster={data.preview.poster}>
        <source src={webm} type="video/webm" />
        <source src={mp4} type="video/mp4" />
    </video>
);
```

This also goes for all the images on the site too - so that can either be part of this pass, or part of the polish pass, as I think there's an argument for both - your call!

pookage: if you stick to the "low"/"medium"/"high" names as well, you could create a util function to reduce repetition of this - just like:

```
const { webm, mp4 } = getBestSource(data.preview.sources);
```

src/components/PreviewOverlay/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+	.details {
+		z-index: 2;
+	}
+
+		.episode {
+			font-size: 1em;
+		}
+
+		.title {
+			font-size: 1.46em;
```
pookage: `1.5`? 🤷‍♂

src/components/PreviewOverlay/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+
+			.button {
+				display: flex;
+				align-items: center;
+
+				&::before, 
+				&::after {
+					display: inline-block;
+					width: 0.8em;
+					height: 1.1em;
```
pookage: I feel like this is absolutely a situation in which an aspect-ratio would help...

src/components/PreviewOverlay/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+
+			.icon {
+				position: relative;
+				height: 100%;
+				width: 100%;
+			}
+
+				.bar {						
+					width: 100%;
+					height: 5px;
```
pookage: just checking - is `px` what we mean here?

src/index.js
```diff
@@ -1,14 +1,17 @@
 import React from "react";
 import ReactDOM from "react-dom";
-import App from "./components/App/";
+import App from "COMPONENTS/App/";
+import { ClientProvider } from "COMPONENTS/Client/";
```
pookage: Lets' not go back and refactor this, but I think that, for other projects, `import { Provider as ClientProvider } from "COMPONENTS/Client/"` is the way to go here. That way every context exports the same thing - `{ Context, Provider }`. It feels more reliable and sturdy.

src/components/EpisodePreview/EpisodePreview.jsx
```diff
@@ -64,92 +50,57 @@ export default function EpisodePreview(props){
...
+				>
+					<img
+						className={ s.image }
+						src={ thumbSrc }
+						alt={ thumbDesc }
+					/>
+					<div className={ s.icon }>
+						<PlayIcon 
+							colour="#FFF" 
 						/>
```
pookage: Ah, almost - this isn't quite what I meant- howcome we should just be able to do 
```
<PlayIcon
    className={s.icon}
    colour="'#FFF"
/>
```

src/components/App/App.jsx
```diff
@@ -13,6 +17,46 @@ import { s } from "./";
...
+
+	useEffect(updateResizeListeners, []);
+
+
+	function updateResizeListeners(){
+		window.addEventListener("resize", syncViewSize);
+		return () => {
+			window.removeEventListener("resize", syncView)
+		};
+	}// updateResizeListeners
```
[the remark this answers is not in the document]
fshenton: All good, I think I must have ordered something wrong so it was looking for dispatch before dispatch was declared 🤔 

I know i'd had problems with context before, so I probably just assumed it was the same thing.

src/components/Client/index.js
```diff
@@ -0,0 +1,5 @@
+import reducer from "./reducer.js";
+import * as ACTIONS from "./actions.js";
+import { ClientContext, ClientProvider } from "./context.jsx";
+
+export { reducer, ACTIONS, ClientContext, ClientProvider };
```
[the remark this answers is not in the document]
fshenton: That would work, but honestly I see both context and provider having equal usefulness depending on the component that needs Client's care and support.

src/components/Client/reducer.js
```diff
@@ -0,0 +1,37 @@
...
+			if(!value){
+				const dimensions = UTILS.getDeviceDimensions(); // currently just using width
+				width = dimensions["width"];
+			}	
+			else {
+				width = value;
+			}
+			
+			const tabletStart = 768;
+			const desktopStart = 1200;
```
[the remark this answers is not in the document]
fshenton: 👌

src/components/EpisodePreview/styles.scss
```diff
@@ -1,5 +1,102 @@
...
+		width: 50%;
+		padding-top: 10%;
+		--thumbnail_height: 50vh;
+	}
+}
+
+//Desktop fullscreen
+@media (min-width: 1200px) {
+	.wrapper {
+		position: initial;
```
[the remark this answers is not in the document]
fshenton: 😏

src/components/EpisodesNav/renderUtils.jsx
```diff
@@ -37,10 +56,39 @@ function renderPaginationLink(data, index){
...
+	}
+	
+	const offset = index - activeIndex;
+	
+	let axis, factor;
+	if(isLarge){
+		axis   = "Y";
+		factor = 1.5;
+	}
+	else{
```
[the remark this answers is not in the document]
fshenton: Feels a bit crammed this way, but i do dig the `} else {` form. tis quite sensual

src/components/PreviewOverlay/PreviewOverlay.jsx
fshenton: 🙍

src/components/PreviewOverlay/PreviewOverlay.jsx
fshenton: 🙍‍♂

src/components/PreviewOverlay/styles.scss
fshenton: yeah it's just copied from the navmenu project. Placeholder for this phase

src/components/PreviewOverlay/PreviewOverlay.jsx
fshenton: I'll add to the board, doesn't seem appropriate for layout phase :)

src/components/PreviewOverlay/PreviewOverlay.jsx
fshenton: I've redone the functions, the rest of the things being passed as props were themselves passed as props from Episodes -> Episode -> PreviewThumb, so we're just continuing the way we are really 🤷‍♂

src/components/EpisodePreview/EpisodePreview.jsx
fshenton: that doesn't actually do anything, unless we unpack the classname in the Play component and then use that as the classname. Which is what we had before.

The classname for the SVG is `icons icon`, giving the PlayIcon a className just passed that as a prop, it doesn't actually allow us to select it in our PreviewOverlay styling.

fshenton:
@pookage I'm very confused as to why the fade + episode nav are visible over the top of the PreviewOverlay ONLY on landscape orientations 🤔 any ideas?

pookage:
So, after some investigating - `PreviewThumb__wrapper` has `position:fixed`, and so creates a new stacking context at that point without defining a new `z-index`; with no `z-index` defined it will default to 'auto', which is basically `0`.

Now, the fade that's overlapping is a child of `Episode__content` - a sibling of `PreviewThumb__wrapper` - which does not have a `position` property defined and so _is not stacking context at all_ - SO, when `Episode__fade` has `position: fixed` defined - _it becomes a sibling stacking context to `PreviewThumb_wrapper`_ and the higher z-index wins.  As `PreviewThumb__wrapper` has no `z-index` declared, and `Episode__fade` has `z-index: 2` - 2 > auto, and the fade overlaps the `PreviewThumb__wrapper`.

The solution? `z-index: 3` on `PreviewThumb__wrapper`.

For the sake of tidiness, this can also mean that `PreviewOverlay__wrapper` can remove its `z-index`, as it's not fighting with any other stacking-context siblings and so the default `auto` is fine 👍

PR #49, opened 2019-10-17
Feature/layout/static elements

fshenton:
- Left navigation positioning as we must have chosen that spacing for a reason......right? 😟 
- Did some funky selective rendering for all the static elements
- Probably a bit blunt force with sizing and spacing values


src/components/HomePermalink/HomePermalink.jsx
```diff
@@ -1,11 +1,28 @@
...
+	const {
+		open
+	} = useContext(Nav).state;
+
+	//RENDER
+	//----------------------------
+
+	const isVisible = !open;
+
 	return (
```
pookage: rather than do the check inside the return, it makes more sense to either return something or not, y'know?  I'd say do

```
if(isVisible) return (<Link ....... />);
else return null;
```

pookage: might not need the `else`? either `react` or `preact` gets grumpy, if you don't explicitly return something, but I can't remember which 🤦‍♂

src/components/Socials/Socials.jsx
```diff
@@ -14,9 +19,12 @@ export default function Socials(){
...
+	const isVisible = !isSmall || open;
+
 	return (
-		<ul className={ `${s.wrapper} ${open ? s.open : s.closed}` }>
-			{ links }
-		</ul>
+		isVisible &&
+			<ul className={ `${s.wrapper} ${open ? s.open : s.closed}` }>
+				{ links }
+			</ul>
```
pookage: same as the other component - as it's the only thing being returned then I'd return different stuff rather than doing conditionals inside the return.

OOH, or, here's a point - maybe these conditional renders should be happening in the parent element instead? That way this component doesn't even need to exist until it's rendered...

src/components/WatchNow/WatchNow.jsx
```diff
@@ -1,10 +1,33 @@
...
 	return (
-		<Link destination="https://www.history.co.uk/shows/vikings">
-			Watch Now
-		</Link>
+		isVisible &&
+			<Link 
+				className={ s.watch }
+				destination="https://www.history.co.uk/shows/vikings">
+				Watch Now
+			</Link>
```
pookage: as above

src/components/WatchNow/styles.scss
```diff
@@ -0,0 +1,13 @@
+.watch {
+	display: flex;
+	position: absolute;
+	width: 11em;
+	height: 3.45em;
```
pookage: is this actually width / height, or is this padding around the text?

fshenton:
I'm going to sort this after Episodes comes back in. Now using Client component for the context, which allows us to selectively render in App because it now knows the device dimensions at that level.

fshenton:
Currently waiting for Episode to be merged into Layout, so i can then sync to get latest, make the changes described in this PR and then bring this branch into Layout also.

src/components/HomePermalink/HomePermalink.jsx
fshenton: react definitely is not please when there's no return! :P

PR #55, opened 2019-10-25
Layout

fshenton:
Notes:
- further tweaks needed for static elements positions/spacing. Will do later
- there's some annoying z-fighting between navigation menu and preview overlay, which i am really struggling to work through. One or the other is always on top and they overlap either when nav is open or the preview overlay is

Otherwise, looking decent.


src/components/WatchNow/styles.scss
```diff
@@ -0,0 +1,25 @@
+// Mobile portrait
+.watch {
+	display: flex;
+	position: absolute;
+	width: 11em;
+	height: 3.45em;
+	bottom: 5%;
+	left: 50%;
+	transform: translateX(-50%);
+	z-index: 3;
```
pookage:
```suggestion
	z-index: 4;
```
Otherwise the button looks like it's behind the ghosts on the Episodes page.

